PR #78, opened 2021-03-19
Ui chat

RoxaneBurri:


app/src/main/java/com/sdp/movemeet/chat/Message.java
```diff
@@ -0,0 +1,48 @@
+package com.sdp.movemeet.chat;
+
+import java.util.Date;
+
+public class Message {
+    private String messageUser;
+    private String messageText;
+    private String messageUserId;
+    private long messageTime;
```
Antho1426, line 9: Why do you use a "long" type here?

app/src/main/java/com/sdp/movemeet/chat/chat_activity.java
```diff
@@ -0,0 +1,16 @@
...
+import android.os.Bundle;
+
+import com.sdp.movemeet.R;
+
+public class chat_activity extends AppCompatActivity {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_chat_activity);
```
Antho1426, line 14: a little empty all that... 😆

app/src/main/res/layout/message_out.xml
```diff
@@ -0,0 +1,94 @@
...
+                        android:textStyle="normal|bold" />
+
+                </LinearLayout>
+
+                <View
+                    android:layout_width="match_parent"
+                    android:layout_height="1dp"
+                    android:background="#26abc2"/>
+
+                <TextView
```
Antho1426, line 56: Do we want to use the "armata" font everywhere (even in the TextViews of the chat), wouldn't it be a good idea?

app/src/main/java/com/sdp/movemeet/chat/Message.java
RoxaneBurri: The time can't be change, it's set at current time when message is create, and so it cast in Date.
It's most easy to get a long, but the part for message in chat is not implemented yet, so it can change if we need.

jlstoupy98 (approved):

Antho1426 (approved):

Antho1426 (approved):

app/src/main/java/com/sdp/movemeet/chat/chat_activity.java
RoxaneBurri: It's your work ;)

app/src/main/res/layout/message_out.xml
RoxaneBurri: Right ! I fix it !

PR #94, opened 2021-03-25
Feature/logged user map

Walfar:
Activities are now retrieved from the database and displayed on the map. 
When a user adds a new activity, the list of activities is updated.

Only the top 2 activities near the user are displayed on the map (using a distance calculator).


app/src/main/java/com/sdp/movemeet/DistanceCalculator.java
```diff
@@ -1,68 +1,130 @@
...
         return sorted;
     }
 
     public Double calculateDistance(Double lat1, Double lng1, Double lat2, Double lng2) {
-        return 2 * EARTH_RADIUS * Math.asin(Math.sqrt(Math.pow(Math.sin((lat2 - lat1)/2),2) + Math.cos(lat1) * Math.cos(lat2) * Math.pow(Math.sin((lng2 - lng1)/2),2)));
+        // Double p = 0.017453292519943295;
+        Double p = Math.PI / 180;
+
+        //Double a =  Math.pow(Math.sin((lat2 - lat1)/2),2) + Math.cos(lat1) * Math.cos(lat2) * Math.pow(Math.sin((lng2 - lng1)/2),2);
+        Double a = 0.5 - Math.cos((lat2 - lat1) * p)/2 +
```
jlstoupy98, line 108: What is the purpose of the two commented lines?

kotleta2007 (approved):

app/src/main/java/com/sdp/movemeet/DistanceCalculator.java
```diff
@@ -1,68 +1,130 @@
...
+import java.util.Comparator;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class DistanceCalculator {
-    private final Double EARTH_RADIUS = 6.3781 * Math.pow(10, 3);
+    // private final Double EARTH_RADIUS = 6.3781 * Math.pow(10, 3);
+
+    // This constant is more ~canonical~
+    private final Double EARTH_RADIUS = 6.371 * Math.pow(10, 3);
```
Antho1426, line 17: Good point to use the formula used by everyone!

app/src/main/java/com/sdp/movemeet/DistanceCalculator.java
```diff
@@ -1,68 +1,130 @@
...
     private Double userLatitude, userLongitude;
-    private ArrayList<Activity> activities = null;
+    private ArrayList<Pair> activityDistanceMap = null;
```
RoxaneBurri, line 19: why do you  use a pair of activity ?

app/src/main/java/com/sdp/movemeet/HomeScreenActivity.java
```diff
@@ -1,26 +1,48 @@
...
     public void signIn(View v) {
         startActivity(new Intent(getApplicationContext(), LoginActivity.class)); // redirecting the user to the "Login" activity
     }
 
     public void noAccount(View v) {
-        startActivity(new Intent(getApplicationContext(), MapsActivity.class)); // redirecting the user to the "Login" activity
+        Intent intent = new Intent(getApplicationContext(), MapsActivity.class); // redirecting the user to the "Login" activity
+        startActivity(intent);
     }
+
```
jlstoupy98, line 46: It looks cleaner in this form, maybe do the same with the previous method?

app/src/main/java/com/sdp/movemeet/HomeScreenActivity.java
```diff
@@ -1,26 +1,48 @@
...
+        updater.updateListActivities();
     }
 
     public void signIn(View v) {
         startActivity(new Intent(getApplicationContext(), LoginActivity.class)); // redirecting the user to the "Login" activity
     }
 
     public void noAccount(View v) {
-        startActivity(new Intent(getApplicationContext(), MapsActivity.class)); // redirecting the user to the "Login" activity
+        Intent intent = new Intent(getApplicationContext(), MapsActivity.class); // redirecting the user to the "Login" activity
```
Antho1426, line 43: False comment

app/src/main/java/com/sdp/movemeet/DistanceCalculator.java
kotleta2007: Good question!

I have used a Pair so that the .sort() method not only sorts the activities, but also saves their distances.

This way, the user can quickly get access to activities in a radius <= N km.

app/src/main/java/com/sdp/movemeet/HomeScreenActivity.java
Walfar: yep, I didn't see that. I'll change that as soon as possible ;)

app/src/main/java/com/sdp/movemeet/HomeScreenActivity.java
Walfar: ok, I will change that

app/src/main/java/com/sdp/movemeet/DistanceCalculator.java
kotleta2007: The first constant was just a hardcoded PI/180.
I decided to replace it with the explicit expression rather than a "magic number" so that it makes more sense.

The second comment was for the previously used, slow and unoptimized formula. It is now useless.

Thanks for the question, this is essentially dead code and I shall remove these comments in my next commit.

app/src/main/java/com/sdp/movemeet/DistanceCalculator.java
kotleta2007: Thanks!

Antho1426 (approved):

PR #104, opened 2021-04-01
set the participants limit

jlstoupy98:
Some adjustments so that the limit of participants when creating an activity is really taken into account.


app/src/main/java/com/sdp/movemeet/Activity/Activity.java
```diff
@@ -277,7 +277,10 @@ public void addParticipantId(String participant){
             throw new IllegalArgumentException();
         }
         if(participantsId.contains(participant)){
-            throw new IllegalArgumentException();
+            throw new IllegalArgumentException("Already Register");
+        }
+        if(participantsId.size() >= this.numberParticipant){
```
Walfar, line 282: The exception should occur when the size of the list is strictly greater than the number of participants, right ? Or do we take in account the organizer, meaning that his id is not in the list but he is still considered as a participant ?

jlstoupy98: If I'm not mistaken, there must be a >=. Indeed, let's imagine a number of participants limited to two, if two people are already registered, we have participantsId.size() == 2. So if a third person tries to register and the size is already the same as the number of people, the registration must fail.

Antho1426 (approved):

RoxaneBurri (approved):

PR #108, opened 2021-04-06
Feature/onclick activities

Walfar:
I changed mapsActivity into two fragments: 1 as a mini-map for the user to set location of the activity when uploading, and 1 as the main fragment when the user accesses the map to see activities around. The user can now click on the map to choose the location of the activity, or directly on the mini-map. Either way, this will set the address of the activity automatically, based on the coordinates. Also, I attempted to modify some tests for the map, but there still are some features that are hard to test, such as the markers on click. I also added a test for the ActivitiesUpdater class.


app/src/androidTest/java/com/sdp/movemeet/ExampleInstrumentedTest.java
```diff
@@ -11,11 +11,7 @@
 import static org.junit.Assert.*;
 
 
-/**
- * Instrumented test, which will execute on an Android device.
- *
- * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
- */
+
```
bitshrine: I believe this whole file can just be deleted... We're somewhat familiar with Instrumentation tests by now

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+   @Rule
+   public FragmentTestRule<?, MainMapFragment> fragmentTestRule =
+           FragmentTestRule.create(MainMapFragment.class);
+
+
+    @Before
+    public void setUp() throws InterruptedException {
+        fAuth = FirebaseAuth.getInstance();
+        fAuth.signInWithEmailAndPassword("victor.carles@epfl.ch", "1234567890").addOnCompleteListener(new OnCompleteListener<AuthResult>() {
```
bitshrine: We should formalize the credentials we use to log in during tests, so that it's all tied to a single account. It's probably better to avoid keeping several dummy users in the database, pending a more secure way of testing scenarios that require authentication. I used "test@test.com" with the password "password" for instance

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+
+    @Test
+    public void mainMapFragment_isDisplayed() throws InterruptedException {
+        onView(withId(R.id.fragment_map)).check(matches((isDisplayed())));
+    }
+
+    @Test
+    public void mainMapFragment_userClickingOnMapAddsNewActivity() throws UiObjectNotFoundException, InterruptedException {
```
bitshrine, line 99: Have you looked into using Mockito to intercept callbacks and function calls? That might help with testing markers and user clicks

app/src/main/java/com/sdp/movemeet/map/MiniMapFragment.java
```diff
@@ -0,0 +1,85 @@
...
+
+        supportMapFragment = (SupportMapFragment) getChildFragmentManager().findFragmentById(R.id.google_map);
+
+        fusedLocationProviderClient = LocationServices.getFusedLocationProviderClient(supportMapFragment.getActivity());
+        fetchLastLocation();
+
+        return view;
+    }
+
+    //TODO refactor this ?
```
bitshrine: If we can guarantee that all actions requiring permissions happen after a permission check, we might be able to refactor these permission request code blocks into functions and then suppress warnings for permission request requirements, but that seems potentially dangerous

app/src/main/java/com/sdp/movemeet/MainActivity.java
```diff
@@ -51,7 +54,7 @@ public void handleRegisterUser() {
     }
 
     public void goToMaps(View view) {
-        startActivity(new Intent(this, MapsActivity.class));
+        startActivity(new Intent(getApplicationContext(), MapsActivity.class));
```
bitshrine: What's the difference between using `this` and `getApplicationContext()` here?

app/src/androidTest/java/com/sdp/movemeet/ExampleInstrumentedTest.java
Walfar: yep, indeed !

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
Walfar: Ok, good to know ! I will use those credentials from now on

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
Walfar: The problem I have with this method is that, for some reason, the UIAutomator doesn't click on the map, although it should. At first, I thought it was because the user was not logged in (hence not being able to add an activity by clicking on the map), but I first check that he is logged with the assertNotNull(user), so the problem comes from the UIdevice. Do you think I could find a way to test the onMapClick with Mockito ? I've never used it, maybe I should look deeper on it.

app/src/main/java/com/sdp/movemeet/map/MiniMapFragment.java
Walfar: I made this comment mostly because I'm using the exact same blocks in two different classes (the two fragments for the map). I thought we could maybe create an abstract class used for the permission checks, to satisfy code climate. I didn't have the time to dive into this matter, but it shouldn't be that long.

app/src/main/java/com/sdp/movemeet/MainActivity.java
Walfar: Yep, you're right. I could simply use "this" haha

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
bitshrine: Ohhh, I see. Mockito could still be used to inject your own test logic in a callback, but I don't think it could help with clicking. Maybe there's some way to make the map focused so that all clicks and interactions go through the map first?

app/src/main/java/com/sdp/movemeet/map/MiniMapFragment.java
bitshrine: Oh! A whole class might be something to look into indeed

app/src/main/java/com/sdp/movemeet/MainActivity.java
jlstoupy98: all interactions related to navigation have been moved to a navigation class to be accessible from the menu and avoid redundancy on all pages. This will be modified when you merge with the main branch

README.md
```diff
@@ -3,3 +3,14 @@
 ### SDP spring semester 2021
```
jlstoupy98: Don't forget to remove your sprint report from the readme if it has been put in the wiki.

app/src/androidTest/java/com/sdp/movemeet/Login/LoginActivityTest.java
```diff
@@ -103,5 +103,5 @@ public void Login_TrueNonEmpty(){
         // Start the screen of your activity.
         onView(withId(R.id.nav_view)).perform(NavigationViewActions.navigateTo(R.id.nav_logout));
```
Antho1426, line 104: So some tests of the LoginActivityTest don't pass and you commented them out? I guess JL made most of those tests work in his last pull request.

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Rule
+    public GrantPermissionRule mRuntimePermissionRule = GrantPermissionRule.grant(Manifest.permission.ACCESS_FINE_LOCATION);
+
+   @Rule
+   public FragmentTestRule<?, MainMapFragment> fragmentTestRule =
+           FragmentTestRule.create(MainMapFragment.class);
+
+
+    @Before
+    public void setUp() throws InterruptedException {
```
Antho1426, line 73: If this test works this way, that's a clever way to authenticate the user without having to go through all the previous activities from the beginning of the app flow (LoginActivity, MainActivity, etc.)! Well done! 👍

app/src/main/java/com/sdp/movemeet/HomeScreenActivity.java
```diff
@@ -1,54 +1,41 @@
...
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_home_screen);
+
         ActivitiesUpdater updater = ActivitiesUpdater.getInstance();
         updater.updateListActivities();
     }
 
     public void signIn(View v) {
-        startActivity(new Intent(getApplicationContext(), LoginActivity.class)); // redirecting the user to the "Login" activity
+        //if (user != null) startActivity(new Intent(this, MainActivity.class));
```
Antho1426, line 30: We can remove this commented line in a future commit.

app/src/main/java/com/sdp/movemeet/HomeScreenActivity.java
```diff
@@ -1,54 +1,41 @@
...
+        startActivity(new Intent(this, LoginActivity.class)); // redirecting the user to the "Login" activity
```
Antho1426, line 31: I made this comment the very first week, but it is as well not very useful, we can remove it.

app/src/main/java/com/sdp/movemeet/UploadActivityActivity.java
```diff
@@ -344,7 +363,6 @@ private Activity validateActivity() {
         validDate = true;
 
         String organizerId = FirebaseAuth.getInstance().getCurrentUser().getUid();
-
         Activity activity = new Activity(
```
Antho1426, line 366: Do you think it will be the right place to also create the "ActivityChatID" for identifying the chat corresponding to this specific sport activity? I'm interested in this since that's certainly what I'm going to work on next week ^^

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
```diff
@@ -0,0 +1,208 @@
...
+        ActivitiesUpdater updater = ActivitiesUpdater.getInstance();
+        activities = updater.getActivities();
+
+        user = fAuth.getCurrentUser();
+
+        return view;
+    }
+
+   /*
+    @SuppressWarnings("deprecation")
```
Antho1426, line 87: I guess this was to grant the permission to fetch the user location? Why did you remove this part of code? Do we have a moment where a pop up message appears and asks the user for letting us retrieve his location?

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
```diff
@@ -0,0 +1,208 @@
...
+        newActivityMarker.remove();
+        Intent intent = new Intent(supportMapFragment.getActivity(), UploadActivityActivity.class);
+        //use Bundle to pass latlng instance to intent
+        Bundle arg = new Bundle();
+        arg.putParcelable("position",marker.getPosition());
+        intent.putExtra("bundle", arg);
+        //remove marker
+        startActivity(intent);
+    }
+}
```
Antho1426, line 208: Very good job by the way! It's a clear file and we can easily understand the main lines by going through it even if we don't work on the map features! Bravo!

app/src/main/java/com/sdp/movemeet/map/MiniMapFragment.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.android.gms.maps.SupportMapFragment;
+import com.google.android.gms.maps.model.LatLng;
+import com.google.android.gms.maps.model.Marker;
+import com.google.android.gms.maps.model.MarkerOptions;
+import com.google.android.gms.tasks.Task;
+import com.sdp.movemeet.R;
+import com.sdp.movemeet.UploadActivityActivity;
+import com.sdp.movemeet.utility.LocationFetcher;
+
+public class MiniMapFragment extends Fragment implements OnMapReadyCallback, GoogleMap.OnMapClickListener {
```
Antho1426, line 30: What is actually the difference between MiniMapFragment and the other main map fragment? Does MiniMapFragment only show the nearby activities, but we can't create a new activity directly from it?

app/src/androidTest/java/com/sdp/movemeet/Login/LoginActivityTest.java
Walfar: Those tests were already commented in the main, we'll have to take a look on it

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
Walfar: Yes, you can just log in this way ;)

app/src/main/java/com/sdp/movemeet/HomeScreenActivity.java
Walfar: I left it here because I felt we had to add something similar at some point, i.e if the user is already logged in, there is no need to bring him to the LoginActivity. It would just be cumbersome for the user. We should directly bring him to the main activity. I commented it because it made the tests fail, but I think it would be important to think about it at some point.

app/src/androidTest/java/com/sdp/movemeet/Login/LoginActivityTest.java
Antho1426: Effectively, it didn't pass, but we keep it for the future!

app/src/main/java/com/sdp/movemeet/UploadActivityActivity.java
Walfar: Yeah sure !

Antho1426: Yes, it will be here! I will create an additional field for the "ActivityChatID" then! 👍

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
Walfar: In the LocationFetcher,when permission is not granted, the permissions are requested. This method should just be used for additional functionalities onRequest, but I think it should be worked in the activity and not in the fragment (hence deprecated). I will take a further look on it next week.

app/src/main/java/com/sdp/movemeet/map/MiniMapFragment.java
Walfar: The mini map is used in the UploadActivity to let the user choose the location on click. The principal difference is that it doesn't display the activities, and directly sets the address of the location clicked. Whereas the main fragment is displayed in the main activity, to display all activities and with the possibility for the user to go to the UploadActivity fragment by clicking on where he wants to set his activity. It's quite similar in the idea, but still have different behaviors, hence the necessity to use two different classes.

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
Walfar: thank you !

Antho1426 (approved):

app/src/androidTest/java/com/sdp/movemeet/Activity/UploadActivityActivityTest.java
```diff
@@ -72,6 +72,11 @@ public static ViewAction forceDoubleClick() {
         };
     }
 
+    @Test
+    public void do_nothing() {
```
jlstoupy98, line 76: Why is there this empty test?

app/src/main/java/com/sdp/movemeet/Activity/Activity.java
```diff
@@ -71,7 +71,7 @@ public Activity(String activityId,
         if(activityId == null || organizerId == null || title == null || numberParticipant <= 0 )
             throw new IllegalArgumentException();
 
-        if(description == null || date == null || duration <= 0 || sport == null || address == null)
+        if(date == null || sport == null || address == null || duration <= 0 || description == null)
```
jlstoupy98, line 74: Why did you change the order of the arguments in the if ?

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
```diff
@@ -0,0 +1,208 @@
...
+
+    private static final String TAG = "Maps TAG";
+
+    public static MainMapFragment newInstance() {
+        return new MainMapFragment();
+    }
+
+    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
+
+        View view = inflater.inflate(R.layout.fragment_map, container, false);
```
jlstoupy98, line 71: Well done! I couldn't get your fragment in and it must have been from that "inflate"

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
```diff
@@ -28,7 +31,7 @@
 
     private static final String TAG = "Activities updater TAG";
 
-    public static final int MAX_NUMBER_ACTIVITIES = 10;
+    public static final int MAX_NUMBER_ACTIVITIES_TO_DISPLAY = 10;
```
jlstoupy98, line 34: This value is fixed for the moment but it should vary according to the choice of the user to filter the closest activity, right?

app/src/androidTest/java/com/sdp/movemeet/Activity/UploadActivityActivityTest.java
Walfar: I added this test when I commented the rest of the tests so that this class still runs, but I should delete it now indeed !

app/src/main/java/com/sdp/movemeet/Activity/Activity.java
Walfar: I tried to delete some fields during tests, and added it back again.

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
Walfar: Yes indeed !

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
Walfar: Exactly

Walfar: The updater is not optimized yet, so displaying too much activities on the map makes the app crash. That's why I set up a threshold, but in the near future I'll optimize it so that we add all activities to the list. The distance filtering that Mark implemented will then be able to only display closest activities indeed.

jlstoupy98 (approved):
Good job !

PR #133, opened 2021-04-15
added sports + createdAt attribute to activity (for ordering)

Walfar:


app/src/androidTest/java/com/sdp/movemeet/Activity/ActivityNotNullDescriptionTest.java
```diff
@@ -50,7 +50,8 @@
             DUMMY_DATE,
             DUMMY_DURATION,
             DUMMY_SPORT,
-            DUMMY_ADDRESS
+            DUMMY_ADDRESS,
+            DUMMY_DATE
```
bitshrine, line 54: Doesn't this conflict with the DUMMY_DATE at line 50?

app/src/test/java/com/sdp/movemeet/Activity/ActivityTest.java
```diff
@@ -37,7 +37,8 @@ public static Activity createFakeActivity() {
                 DUMMY_DATE,
                 DUMMY_DURATION,
                 DUMMY_SPORT,
-                DUMMY_ADDRESS
+                DUMMY_ADDRESS,
+                DUMMY_DATE
```
bitshrine: I feel like using DATE for both the creation date and activity date might cause conflicts or at the very least confusion.

app/src/androidTest/java/com/sdp/movemeet/Profile/HomeScreenActivityToUserProfileActivity.java
```diff
@@ -47,6 +49,9 @@
     @Rule
     public ActivityTestRule<HomeScreenActivity> mActivityTestRule = new ActivityTestRule<>(HomeScreenActivity.class);
 
+    @Rule
+    public GrantPermissionRule mRuntimePermissionRule = GrantPermissionRule.grant(Manifest.permission.ACCESS_FINE_LOCATION);
+
     @Test
     public void homeScreenActivityToUserProfileActivity() {
         /*onView(withId(R.id.signInButton)).perform(click());
```
bitshrine, line 55: If this test fails because of the login part, the best idea is to sign in programmatically using `FirebaseAuth` and launching the activity afterwards using `ActivityScenario.launch()` (this would mean not using the `ActivityTestRule`)

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
```diff
@@ -95,6 +124,92 @@ public void mainMapFragment_isDisplayed() throws InterruptedException {
...
+                        } catch (InterruptedException e) {
+                            e.printStackTrace();
+                        }
+                        List<Activity> activities = updater.getActivities();
+                        Activity act_in_collection = activities.get(activities.size()-1);
+                        assertEquals(act.getActivityId(), act_in_collection.getActivityId());
+                        assertEquals(act.getAddress(), act_in_collection.getAddress());
+                        assertEquals(act.getDate(), act_in_collection.getDate());
+                        assertEquals(act.getDescription(), act_in_collection.getDescription());
+                        //assertEquals(act.getLatitude(), act_in_collection.getLatitude());
```
bitshrine: I think these issues might be caused by weird ordering of the map loading / location fetching when handling `MapFragments`.

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
```diff
@@ -111,7 +96,7 @@ public void onMapReady(GoogleMap googleMap) {
         LatLng userLatLng = new LatLng(currentLocation.getLatitude(), currentLocation.getLongitude());
         MarkerOptions markerOptions = new MarkerOptions().position(userLatLng).title("I am here !");
         googleMap.animateCamera(CameraUpdateFactory.newLatLng(userLatLng));
-        googleMap.animateCamera(CameraUpdateFactory.newLatLngZoom(userLatLng, 12.0f));
+        googleMap.animateCamera(CameraUpdateFactory.newLatLngZoom(userLatLng, 15.0f));
```
bitshrine: Should this constant be moved to a public static attribute? Maybe we should have a uniform zoom on map fragments.

app/src/main/java/com/sdp/movemeet/utility/LocationFetcher.java
```diff
@@ -7,40 +7,39 @@
...
-import androidx.fragment.app.Fragment;
 
 import com.google.android.gms.location.FusedLocationProviderClient;
 import com.google.android.gms.maps.OnMapReadyCallback;
 import com.google.android.gms.maps.SupportMapFragment;
 import com.google.android.gms.tasks.OnSuccessListener;
 import com.google.android.gms.tasks.Task;
 
-public class LocationFetcher {
+public abstract class LocationFetcher {
```
bitshrine, line 17: It seems like this class's functionality relies on asynchronous operations but the results we want can be accessed synchronously. I think this has caused issues in tests when retrieving the location for instance, as the class returns null since the operation has not yet completed. We should probably refactor this class such that we can wait for the results to succeed or fail before being able to use them.

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
```diff
@@ -48,91 +58,121 @@ public static ActivitiesUpdater getInstance() {
...
+        if (obj == null) return 0;
+        else return obj.intValue();
+    }
+
+    private double convertObjToDouble(Object obj) {
+        if (obj == null) return 0;
+        else return (double) obj;
+    }
 
+    private Date getDateObj(Date date) {
```
bitshrine, line 158: You might be able to make these conversion functions into a single one using generics

bitshrine (changes requested):

app/src/androidTest/java/com/sdp/movemeet/Activity/ActivityNotNullDescriptionTest.java
Walfar: Since all we're interested in is the value of the field, I don't think that it is particularly conflicting to have two times the same value. But I agree that it can lead to some confusion, so I've changed the last constant to "DUMMY_CREATION_DATE" for that matter.

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
Walfar: Yes good idea, I've changed it as a static constant called ZOOM_VALUE, used in both the MainMap and the MiniMap

app/src/androidTest/java/com/sdp/movemeet/map/MainMapFragmentTest.java
Walfar: In the previous version of my code, the currentLocation was not initialized before calling the LocationFetcher, hence it could have a null value indeed. I've added a "defaultLocation" that is always assigned to the currentLocation on creation of the fragment to prevent that from happening.

app/src/androidTest/java/com/sdp/movemeet/Profile/HomeScreenActivityToUserProfileActivity.java
Walfar: I don't really looked into this test, since it was already commented when I merged it with my code, so I just assumed it didn't work. But I'll try to follow your advise, and see if it works now !

app/src/androidTest/java/com/sdp/movemeet/utility/ActivitiesUpdaterTest.java
```diff
@@ -45,47 +46,4 @@ public void instanceIsNeverNull() {
         assertNotNull(updater);
     }
 
-    @Test
```
Antho1426, line 48: So you deleted this test to move it to MainMapFragmentTest? But then is this ActivitiesUpdaterTest still useful since it is almost empty ?

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
Walfar: That's right ! I didn't think about it, I will try to do that

app/src/main/java/com/sdp/movemeet/utility/LocationFetcher.java
Walfar: In the previous version of my code, the currentLocation was not initialized before calling the LocationFetcher, hence it could have a null value indeed. I've added a "defaultLocation" that is always assigned to the currentLocation on creation of the fragment to prevent that from happening.

app/src/main/AndroidManifest.xml
```diff
@@ -5,7 +5,10 @@
 
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
Antho1426, line 8: I guess this is to check if the user is connected to internet or not 👍

app/src/main/java/com/sdp/movemeet/Sport.java
```diff
@@ -5,5 +5,17 @@
     Soccer,
     Swimming,
     Tennis,
-    Badminton
+    Badminton,
```
Antho1426, line 8: Wahoo, you added a lot of new sport activities! Do they all have their own icon?

app/src/main/java/com/sdp/movemeet/utility/ActivityCache.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.ArrayList;
+
+import static android.content.Context.MODE_PRIVATE;
+
+public class ActivityCache extends AppCompatActivity {
+
+    public void saveActivitiesInCache(ArrayList<Activity> activities) {
+        ObjectOutput out;
+        try {
+            File outFile = new File(Environment.getExternalStorageDirectory(), "cached_activities");
```
Antho1426: Simple, efficient, nice! Good if local cache works this way in that easy manner!

app/src/main/java/com/sdp/movemeet/Sport.java
Antho1426: Sorry I initially didn't that you also committed a png for each of those new sport activities 👍

app/src/androidTest/java/com/sdp/movemeet/utility/ActivitiesUpdaterTest.java
Walfar: I intend to add some other tests in this class

app/src/main/AndroidManifest.xml
Walfar: That's right. But in the end, I might not use this way of checking the internet connection since we can simply a ping command at runtime to some website (e.g google.com). This might be an easier way to do it, so I might simply delete this line.

app/src/main/java/com/sdp/movemeet/Sport.java
Walfar: Yes ! But I noticed that they all had different sizes, so I had to resize them to 24x24. Some of them lose in quality, so we might have to change these icons later on.

app/src/main/java/com/sdp/movemeet/utility/ActivityCache.java
Walfar: Thank you, but unfortunately it isn't working yet :(
I think the problem comes from the permissions rather than the way the data is stored locally.

PR #155, opened 2021-04-16
Refactor the code to access the backend in a generic way

bitshrine:


app/src/main/java/com/sdp/movemeet/Backend/Firebase/FirebaseDB/FirebaseDBManager.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.firebase.database.DatabaseReference;
+import com.google.firebase.database.FirebaseDatabase;
+import com.google.firebase.firestore.DocumentReference;
+import com.sdp.movemeet.Backend.BackendStorage;
+import com.sdp.movemeet.Backend.Firebase.FirebaseObject;
+import com.sdp.movemeet.Backend.Serialization.BackendSerializer;
+
+import java.util.Map;
+
+abstract class FirebaseDBManager<T extends FirebaseObject> implements BackendStorage<T> {
```
RoxaneBurri, line 13: Really nice interface for Firebase !

app/src/main/java/com/sdp/movemeet/Backend/Firebase/Firestore/FirestoreActivityManager.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public FirestoreActivityManager(FirebaseFirestore db, String collection, BackendSerializer<Activity> serializer) {
+        super(db, collection, serializer);
+        this.db = db;
+        this.collection = collection;
+    }
+
+    @Override
+    public Task<QuerySnapshot> search(String field, Object value) {
+        // TODO: return 50 closest activities and execute query on those
```
RoxaneBurri, line 25: Can the user choose how many closest activities ?

RoxaneBurri (approved):

app/src/main/java/com/sdp/movemeet/Backend/Firebase/Firestore/FirestoreActivityManager.java
bitshrine: Basically, my thinking is that we should only programmatically fetch activities that are close to the user, to avoid having the device fetch all activities in the database and then sorting them locally. For instance, if a users asks for the activities in a 10km radius to their location, the backend service would first fetch 50/100/200 closest activities and check which among those are inside the radius. Of course, we could also fetch all activities inside the radius instead of a fixed number of them, provided we are able to query that to the backend easily. My concern is that all activities are always fetched as of now, which does not seem like a scalable solution.

Walfar: I agree, but I don't understand why this should happen in the search method. Shouldn't there be another method that sorts the db to only retrieve the closest activities, using the distance calculator ? I guess that's how you want to do it. Also, the way the updater updates the list of activites is by sorting them by creation date, and only taking the newly added activities. I'm not sure to see how we can do this "double" sorting (by distance and only new activities) efficiently. But maybe that's a matter for another time ;)

app/src/main/java/com/sdp/movemeet/Activity/Activity.java
```diff
@@ -335,6 +334,15 @@ public DocumentReference setBackendRef(DocumentReference newRef) throws IllegalA
         return backendRef;
     }
 
+    public String getDocumentPath() {
+        return this.documentPath;
+    }
+
+    public String setDocumentPath(String path) {
```
Walfar, line 341: How would we typically define the path ? Do you have an example of value it should take ?

app/src/main/java/com/sdp/movemeet/Backend/Serialization/ActivitySerializer.java
```diff
@@ -50,9 +50,14 @@ public static Activity deserialize(Map<String, Object> data) {
                 (String) data.get(ADDRESS_KEY),
                 (Date) data.get(CREATION_KEY)
         );
+
+        String path = (String) data.get(DOCUMENT_PATH_KEY);
+        if (path != null) act.setDocumentPath(path);
```
Walfar, line 55: Is it problematic if the path is null ? What would happen then ? And how would that be possible ?

app/src/test/java/com/sdp/movemeet/Backend/Firebase/FirebaseDB/FirebaseDBActivityManagerTest.java
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+        @Override
+        public boolean isCanceled() {
+            return false;
+        }
+
+        @Override
+        public Void getResult() {
+            return null;
```
Walfar, line 60: Are all those methods really supposed to return null, or is it just not implemented yet ?

app/src/main/java/com/sdp/movemeet/Backend/Serialization/BackendSerializer.java
```diff
@@ -0,0 +1,9 @@
+package com.sdp.movemeet.Backend.Serialization;
+
+import java.util.Map;
+
+public interface BackendSerializer<T> {
```
Walfar, line 5: Good idea using generic interfaces in case we need to serialize other types of objects :)

Walfar (approved):

app/src/main/java/com/sdp/movemeet/Backend/Firebase/Firestore/FirestoreActivityManager.java
Antho1426: Yep, I agree with Kepler as well.

Concerning filtering activities based on both creation date and distance, I guess this should be feasible, we just have to choose what is the most important criterion (between creation date and distance). I would say filtering activities that are in our radius first and then extracting only the activities that are not finished yet (activity date > current date). Finally we filter out those activities such that they don’t exceed the maximum number of closest activities (50/100/200). In summary we would have 3 passes:
1. Radius
2. Activity date
3. Limit number of activities to display

app/src/main/java/com/sdp/movemeet/Backend/Serialization/ActivitySerializer.java
Antho1426: I guess it would throw a kind of "StorageException" error. At least with Firebase Storage (which is used to store BLOB files on Firebase and is hence not really the same service as Firebase Firestore), this is the kind of error I get when the user access his profile in "UserProfile" because, initially, he has no profile picture and the path leading to this picture hence doesn't exist yet.

Walfar: Ok I see, but then maybe it would be a good idea to catch this exception ? If we check if the path is not null, I think it would be a good idea to have an "else" case as well

app/src/main/java/com/sdp/movemeet/Activity/Activity.java
bitshrine: It would be set on Activity upload, you can call `.getDocumentPath()` once you have the `DocumentReference`

PR #160, opened 2021-04-22
Unregister home

RoxaneBurri:


app/src/main/java/com/sdp/movemeet/HomeUnregister.java
```diff
@@ -0,0 +1,38 @@
...
+    TextView textView;
+    
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_home_unregister);
+
+        createDrawer();
+    }
```
Antho1426: So, if I understand properly, the difference is that now, instead of arriving on the MainActivity, unregistered users land on the HomeRegister and here they have also a side panel on the left but without any possibility (such as UserProfile, UploadActivity, etc.), is this correct?

Antho1426 (approved):

app/src/main/java/com/sdp/movemeet/HomeUnregister.java
```diff
@@ -0,0 +1,38 @@
...
+    Toolbar toolbar;
+    TextView textView;
+    
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_home_unregister);
+
+        createDrawer();
```
jlstoupy98: Is it always necessary to have drawer if user is not logged ? Maybe a simple backspace button would be enough?

app/src/main/java/com/sdp/movemeet/HomeUnregister.java
```diff
@@ -0,0 +1,38 @@
+package com.sdp.movemeet;
```
Walfar: It is a detail, but shouldn't this class be called "MainActivityUnregistered", or something like that ? From my understanding, this isn't the home, but rather the main activity but without the user  menu

app/src/main/java/com/sdp/movemeet/HomeUnregister.java
RoxaneBurri: Yes, there is also no navigation view, because there are only feature for register user

app/src/main/java/com/sdp/movemeet/HomeUnregister.java
RoxaneBurri: Yes you right

app/src/main/java/com/sdp/movemeet/HomeUnregister.java
RoxaneBurri: I can rename the class !

jlstoupy98 (approved):

PR #185, opened 2021-04-27
Unregister activity description

RoxaneBurri:


app/src/main/AndroidManifest.xml
```diff
@@ -64,7 +64,7 @@
                 android:value="com.sdp.movemeet.HomeScreenActivity" />
         </activity>
         <activity android:name=".MainActivity" />
-        <activity android:name=".MainUnregister" />
+        <activity android:name=".MainUnregisterActivity" />
```
Antho1426, line 67: Yes, that's a better name for an Android Activity!

app/src/main/java/com/sdp/movemeet/Activity/ActivityDescriptionActivity.java
```diff
@@ -351,28 +303,11 @@ public void changeActivityPicture(View view) {
...
-        i.putExtra(EXTRA_DESCRIPTION, "1");
-        i.putExtra(EXTRA_TITLE, "1");
-        i.putExtra(String.valueOf(EXTRA_DATE), "1");
-        i.putExtra(String.valueOf(EXTRA_LATITUDE), 1.0);
-        i.putExtra(String.valueOf(EXTRA_LONGITUDE), 1.0);
-        i.putExtra(String.valueOf(EXTRA_NUMBER_PARTICIPANT), 1.0);
-        i.putExtra(String.valueOf(EXTRA_SPORT), "1.0");
-
-        startActivity(i);
-    }*/
```
Antho1426, line 377: Yes, I was hesitating to remove this part as well since it was commented since a while, I will do the same!

app/src/main/java/com/sdp/movemeet/Activity/ActivityDescriptionActivityUnregister.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private void createDurationView() {
+        TextView durationView = (TextView) findViewById(R.id.activity_duration_description);
+        if (act != null) {
+            durationView.setText(String.valueOf((int) act.getDuration()));
+        }
+    }
+
+    private void createAddressView() {
```
Antho1426, line 102: Do we want to display the date of the activity to unregistered (i.e. unverified and potentially malicious) users? For security reasons, I think we should either avoid displaying the date or the address of the activity, what do you think?

app/src/main/java/com/sdp/movemeet/Activity/ActivityDescriptionActivity.java
RoxaneBurri: Yes, it's was for an old task, so we don't need it anymore

app/src/main/java/com/sdp/movemeet/Activity/ActivityDescriptionActivityUnregister.java
RoxaneBurri: Yes, I only remove the organizer name, but we can discuss tonight about it. Jean-Luc create a new layout for this page so I will see it with him.

PR #187, opened 2021-04-29
Change old Firebase Activity-related code to use new generic backend interface

### LINKED ISSUE #179: Replace old Firebase-related Activity code with new generic interface

bitshrine:
closes #179


jlstoupy98:
This is a great job, well done !!!

app/src/main/java/com/sdp/movemeet/Backend/BackendManager.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.google.android.gms.tasks.Task;
+
+import java.util.concurrent.Future;
+
+/**
+ * A generic interface to perform basic storage management
+ * operations with a backend
+ * @param <T> The type of object that this BackendManager will hande
+ */
```
Antho1426: hande --> handle 😉

Antho1426 (approved):

Antho1426:
Yes, looking forward to having this interface, thank you for your tremendous job Kepler!

app/src/main/java/com/sdp/movemeet/models/Activity.java
```diff
@@ -84,7 +84,7 @@ public Activity(String activityId,
         this.longitude = longitude;
         this.latitude = latitude;
         this.description = description;
-        this.documentPath = documentPath;
+        super.documentPath = documentPath;
```
Antho1426, line 87: Why did you use "super" instead of "this" here ? documentPath is already in the constructor ?

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
```diff
@@ -56,7 +57,7 @@
         onView(withId(R.id.button_login)).perform(click());
     }
 
-    @Test public void Login_ShortPassword() throws InterruptedException {
```
kotleta2007, line 59: Why did you remove the `throws InterruptedException` part? Does some other part of the code take care of these exceptions?

app/src/main/java/com/sdp/movemeet/view/navigation/Navigation.java
```diff
@@ -1,55 +1,227 @@
...
+                @Override
+                public void onComplete(@NonNull Task<DocumentSnapshot> task) {
+                    if (task.isSuccessful()) {
+                        User user = serializer.deserialize(task.getResult().getData());
+
+                        fields[0].setText(user.getFullName());
+                        fields[1].setText(user.getEmail());
+                        fields[2].setText(user.getPhoneNumber());
+                        if (fields.length > 3 && user.getDescription() != null && !user.getDescription().isEmpty())
+                            fields[3].setText(user.getDescription());
```
Antho1426, line 151: Filling in the "user description" in the navigation bar wouldn't fill it too much?

app/src/main/java/com/sdp/movemeet/backend/firebase/firebaseDB/FirebaseDBActivityManager.java
```diff
@@ -4,7 +4,7 @@
 import com.sdp.movemeet.models.Activity;
 import com.sdp.movemeet.backend.serialization.BackendSerializer;
 
-public class FirebaseDBActivityManager extends FirebaseDBManager<Activity> {
+public class FirebaseDBActivityManager extends com.sdp.movemeet.backend.firebase.firebaseDB.FirebaseDBManager<Activity> {
```
kotleta2007, line 7: Is this necessary? The full path name is much harder to read

Antho1426 (approved):

app/src/main/java/com/sdp/movemeet/models/FirebaseObject.java
```diff
@@ -3,16 +3,24 @@
 /**
  * An object that can be found in the Firebase Realtime Database/Firestore backend
  */
-public interface FirebaseObject {
```
kotleta2007, line 6: What's the advantage of using an abstract class instead of an interface?

app/src/main/java/com/sdp/movemeet/view/map/GPSRecordingActivity.java
```diff
@@ -112,7 +116,7 @@ public void onMapReady(GoogleMap googleMap) {
     }
 
 
-    public void updatePath(Location location) {
```
kotleta2007, line 115: That's a good point, encapsulation is important

kotleta2007 (approved):

app/src/main/java/com/sdp/movemeet/models/Activity.java
bitshrine: Since `FirebaseObject` is now an abstract class, it has its own `documentPath` attribute which it uses in its default implementation of `getDocumentPath()` and `setDocumentPath()`. Here, I'm setting the document path in the superclass so that the abstract functions can run correctly.

app/src/main/java/com/sdp/movemeet/view/navigation/Navigation.java
bitshrine: I wasn't sure where else this function was meant to be used, so I copied the description part as well so that it may be used elsewhere in the future.

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
bitshrine: For some reason, this test was repeatedly failing locally, and after some trial and error, removing the ` throws InterruptedException` fixed it. If that was a quirk of my machine, we can probably add it back, but Android Studio didn't throw any warnings when I removed it

app/src/main/java/com/sdp/movemeet/backend/firebase/firebaseDB/FirebaseDBActivityManager.java
bitshrine: Android Studio wouldn't recognize it otherwise :(

app/src/main/java/com/sdp/movemeet/models/FirebaseObject.java
bitshrine: WIth an abstract class, we get to have a default implementation for the `getDocumentPath()` and `setDocumentPath()` methods, and as I was implementing FirebaseObjects I realized they were often identical codeblocks across classes. Hence, the change to an abstract class with a single implementation

app/src/main/java/com/sdp/movemeet/models/Activity.java
Antho1426: Okay I see, `Activity.java` is in fact inheriting from (i.e. implementing) `FirebaseObject`!

app/src/main/java/com/sdp/movemeet/view/navigation/Navigation.java
Antho1426: Okay! This will be something new then (but why not)! Because for the moment I guess that only the name, the phone number and the email address of the user (but not his description) are displayed in the navigation bar.

app/src/main/java/com/sdp/movemeet/models/FirebaseObject.java
Antho1426: That's in fact a good point from you Kepler to have changed `FirebaseObject` from interface to an abstract class because you can in fact now create functionality that subclasses (such as User, Activity and Message) can implement or override! On the contrary, an interface only allows you to define functionalities, not implement them.

PR #189, opened 2021-04-29
ActivityDescriptionActivity layout enhancement & patch some memory leaks

jlstoupy98:
I reworked the layout of ActivityDescriptionActivity to make it more presentable.
finish() have been added to avoid overlapping activities each time a startActivity() is called. This way, the activities don't overlap, solving some problems concerning the behavior of the back button


app/build.gradle
```diff
@@ -97,6 +97,7 @@ dependencies {
         exclude module: 'recyclerview-v7'
     }
     annotationProcessor 'androidx.room:room-compiler:2.2.6'
+    implementation 'de.hdodenhof:circleimageview:3.1.0'
```
Antho1426, line 100: What is this new implementation for ?

Antho1426 (approved):

bitshrine (approved):

app/build.gradle
jlstoupy98: it simply adds a tag in the xml that allows to have a circular image. I use it to enhance the design of our layout.

app/src/main/java/com/sdp/movemeet/map/MainMapFragment.java
```diff
@@ -54,7 +50,7 @@
     //current location of the user on the map
     private Location currentLocation;
 
-    private FirebaseAuth fAuth = FirebaseAuth.getInstance();
+    private final FirebaseAuth fAuth = FirebaseAuth.getInstance();
```
RoxaneBurri, line 53: We have to put every fAuth in final

Antho1426: Why is this necessary?

Antho1426 (approved):

PR #227, opened 2021-05-06
Firestore abstraction in Activities Updater

Walfar:
- used firestore abstraction in Activities Updater
- refactored Activities Updater to have better abstraction with listeners and map callbacks
-  small changes on fireAuthentication in tests, to try to solve the Firebase SDK version bug
-  adapted map tests to new abstraction


app/src/androidTest/java/com/sdp/movemeet/viewTest/navigationTest/ProfileActivityNavigationTest.java
```diff
@@ -1,14 +1,15 @@
-package com.sdp.movemeet.Navigation;
+/*package com.sdp.movemeet.viewTest.navigationTest;
```
Antho1426: Yes, the Firebase runloop error might come from one of those Navigation tests. That's a good point to comment all of them to be able to merge on the main branch!

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
```diff
@@ -1,181 +1,89 @@
...
-                Log.d(TAG, e.toString());
+    /**
+     * On success, add the activities from db to the local list of activities
+     *
+     * @param task task containing the requested snapshots, that would then be deserialized into activities
+     */
+    private static Task addActivitiesOnSuccess(Task<QuerySnapshot> task) {
+        return task.addOnSuccessListener(queryDocumentSnapshots -> {
+            for (DocumentSnapshot docSnap : queryDocumentSnapshots.getDocuments()) {
+                activities.add(serializer.deserialize(docSnap.getData()));
```
Antho1426, line 77: Ok I was interested to see how you would use the abstraction "ActivitiesUpdater"! 👍

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
```diff
@@ -1,173 +1,89 @@
...
-        });
+        }).addOnFailureListener(e -> Log.d(TAG, e.toString()));
     }
 
-    public void clearLocalActivities() {
+    /**
+     * Clear the local list of activities
+     */
+    public static void clearLocalActivities() {
         activities.clear();
```
Antho1426, line 87: Is this ".clear()" method also part of the abstraction? Because I'm not able to find it in the BackendManager.java interface...

app/src/main/java/com/sdp/movemeet/backend/firebase/firestore/FirestoreActivityManager.java
```diff
@@ -26,4 +29,23 @@ public FirestoreActivityManager(FirebaseFirestore db, String collection, Backend
         return super.search(field, value);
     }
 
+    /**
+     * Returns a list of recently added activities, with the specified size
+     * @param size number of recently added activities we want to get
+     * @return a querySnapshot for size number of recently added activities
+     */
+    public Task<QuerySnapshot> getRecentlyAddedActivities(int size) {
+        return db.collection(collection).orderBy("createdAt", Query.Direction.DESCENDING).limit(size).get();
```
Antho1426, line 38: Query.Direction.DESCENDING is for listing the activities from the first recent to the oldest, is it correct?

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -84,16 +75,10 @@
...
+    public void setUp() {
         fAuth = FirebaseAuth.getInstance();
-        fAuth.signInWithEmailAndPassword("test@test.com", "password").addOnCompleteListener(new OnCompleteListener<AuthResult>() {
-            @Override
-            public void onComplete(@NonNull Task<AuthResult> task) {
-                user = fAuth.getCurrentUser();
-                uiDevice = UiDevice.getInstance(getInstrumentation());
-            }
-        });
+        fAuth.signInWithEmailAndPassword("test@test.com", "password");
```
jlstoupy98, line 80: Why we didn't need the onCompleteListener anymore ?

app/src/androidTest/java/com/sdp/movemeet/viewTest/navigationTest/ProfileActivityNavigationTest.java
Walfar: Indeed. We will refactor them anyway thanks to Kepler's class

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -217,11 +143,6 @@ public void mainMapFragment_userClickingOnMapAddsNewActivity() throws UiObjectNo
         uiDevice.click(200, 400);
     }
 
-    @After
-    public void after() {
```
jlstoupy98, line 221: Why you remove the logout at the end of the test ? Isn't it necessary for the good progress of all the tests?

app/src/androidTest/java/com/sdp/movemeet/view/navigation/ActivityDescriptionNavigationTest.java
```diff
@@ -15,13 +19,12 @@
 
 public class ActivityDescriptionNavigationTest {
     @Rule
-    public ActivityScenarioRule<HomeScreenActivity> testRule = new ActivityScenarioRule<>(HomeScreenActivity.class);
+    public ActivityScenarioRule<LoginActivity> testRule = new ActivityScenarioRule<>(LoginActivity.class);
```
jlstoupy98, line 22: It's not stupid, you're right to do that, it takes a step away!

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
Walfar: Nothing much changed. But I took the time to also refactor this class, with external listener (instead of internal) and use of the serializer (that I also needed to change to make it work with Firebase).

app/src/androidTest/java/com/sdp/movemeet/view/navigation/security/ActivityDescriptionSecurityTest.java
```diff
@@ -15,6 +18,12 @@
 
 public class ActivityDescriptionSecurityTest {
 
+    @Before
+    public void checkUserNotLogged() {
+        FirebaseAuth firebaseAuth = FirebaseAuth.getInstance();
+        if (firebaseAuth.getCurrentUser() != null) firebaseAuth.signOut();
```
jlstoupy98, line 24: So here, we assume that user is log by default ? Is it not a strange behavior or assumption ?

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
Walfar: No, this is not part of the abstraction as it is used to clear the local activities, not the activities in the database (so it shouldn't be in the abstraction). But since for the moment it is only used once, I might remove it.

app/src/main/java/com/sdp/movemeet/backend/firebase/firestore/FirestoreActivityManager.java
Walfar: Exactly ! And then, I limit(size) to take only the size most recent ones.

jlstoupy98, line 38: It seems smart from an information transfer point of view but what happens if many activities are added to the DB between two "calls" to this function? Not all of them will be retrieved so we have a "hole" in the list of activities, right?

app/src/androidTest/java/com/sdp/movemeet/view/navigation/ActivityDescriptionNavigationTest.java
Walfar: thanks haha

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
Walfar: With this, I had to add a login before every map test, which was redundant as most of those tests don't even need the user to be logged in. Since the following tests are the security tests, that always check that the user is not logged (and if he is, unlogged him), that doesn't create any problem. But we might have to change it  afterwards anyway, when we will mock the authentication.

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
Walfar: The instanciation of the UiDevice is not dependant from the user being logged in, so there is no need to have it in the onCompleteListener. And now that the listener is not doing anything, I could just delete it.

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
Antho1426: Ok, I think I get it!

app/src/main/java/com/sdp/movemeet/backend/firebase/firestore/FirestoreActivityManager.java
Antho1426: Yep right, thank you for your answer!

app/src/androidTest/java/com/sdp/movemeet/view/navigation/security/ActivityDescriptionSecurityTest.java
Walfar: It's not really that we assumer that user is logged, it's simply checking that the user is not logged so that this test has the right behavior. Sure, we could assume that the user is never logged when lauching a test, but this is not a guarantee. As explained above, I mainly did that to fit the fact that I don't logout the user from the map. But I might change this, in which case we could get rid of this check I guess

app/src/main/java/com/sdp/movemeet/backend/firebase/firestore/FirestoreActivityManager.java
Walfar: Good question ! Let's suppose we have a DB of 30 activities, and the local list is empty. On first fetch, the size will be 30 (as in the updateListActivities method, we check the diff size between local list and DB), and this will get all activities. Now, let's suppose meanwhile we have 5 new activites added to the DB. Then, when the user refreshes the map, another diff check is made. Now, the size will be 5 (as the difference from the local list size 30 and DB size 35 is 5). This method will then get the 5 most recent activities, and add them to the local list. I don't see a case where not all activities from DB would be added to the local list.

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
```diff
@@ -1,173 +1,89 @@
...
      */
-    public void updateListActivities(MainMapFragment mapFragment) {
-        this.mapFragment = mapFragment;
-        CollectionReference collection = bam.getActivitiesCollectionReference();
+    public static void updateListActivities(OnCompleteListener listener) {
+        if (listener == null) {
+            Log.d(TAG, "listener is null");
+            return;
+        }
+        Task<QuerySnapshot> allDocTask = firestoreActivityManager.getAll();
```
Antho1426, line 49: Ok here is some abstraction! Nice!

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
```diff
@@ -1,173 +1,89 @@
...
+            //In case there is more activities in local than db (e.g if activities were deleted from db), than we clear the list and refetch
+            //TODO: not an optimized way to do, might be interesting to rethink it
+            if (size < 0) {
+                clearLocalActivities();
+                updateListActivities(listener);
+                return;
             }
+            //Either way, when updating the list, we update the map as well
+            else if (size == 0) allDocTask.addOnCompleteListener(listener);
+            else addActivitiesOnSuccess(firestoreActivityManager.getRecentlyAddedActivities(size)).addOnCompleteListener(listener);;
```
Antho1426, line 65: And this is also one of your other added method for the abstraction! Nice!

Antho1426 (approved):

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
Walfar: thanks !

jlstoupy98 (approved):

PR #240, opened 2021-05-13
Added tests for map's callbacks

Walfar:


app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -63,41 +86,135 @@
...
     }
 
+    @Test
+    public void displayingUserMarkerSetsPositionMarker() {
+        MainMapFragment mapFragment = fragmentTestRule.getFragment();
+        mapFragment.currentLocation = new Location("current location");
+        mapFragment.currentLocation.setLatitude(0);
+        mapFragment.currentLocation.setLongitude(0);
+
+        mapFragment.getActivity().runOnUiThread(new Runnable() {
```
jlstoupy98, line 133: Why do you use runOnUiThread ?

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -63,41 +86,135 @@
...
+            @Override
+            public void run() {
+                mapFragment.displayNearbyMarkers();
+                if (mapFragment.googleMap != null) {
+                    mapFragment.onMarkerClick(mapFragment.activitiesMarkers.get(0));
+                }
+            }
+        });
+        assertNotNull(fAuth.getCurrentUser());
+        if (mapFragment.googleMap != null) intended(hasComponent(ActivityDescriptionActivity.class.getName()));
```
Antho1426, line 177: Why do you need to verify that mapFragment.googleMap is not null here?

jlstoupy98, line 177: Isn't the map supposed to be non-null all the time?

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -63,41 +86,135 @@
 @RunWith(AndroidJUnit4.class)
 public class MainMapFragmentTest {
 
-    private UiDevice uiDevice;
+    private UiDevice uiDevice = UiDevice.getInstance(getInstrumentation());
```
Antho1426, line 89: Why did you change this line to use a specific instance of the uiDevice instead of the uiDevice object itself?

jlstoupy98 (approved):

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
Walfar: The callbacks methods in the map can only run on the main UI thread, so they can't be called directly in the tests. I have to make those calls run on the UI thread, and this is done by using this method !

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
Walfar: Yes, actually the google map is never null when the map methods are called back. But, since Cirrus CI isn't supposed to use an internet connection, the google map would never be loaded (hence would be null), and those tests would fail. I added these checks to make sure that the tests only run under the condition that the google map is loaded. In local, these tests always run correctly, but with Cirrus, they are not doing much unfortunately.

Walfar: (same answer that for JL) Yes, actually the google map is never null when the map methods are called back. But, since Cirrus CI isn't supposed to use an internet connection, the google map would never be loaded (hence would be null), and those tests would fail. I added these checks to make sure that the tests only run under the condition that the google map is loaded. In local, these tests always run correctly, but with Cirrus, they are not doing much unfortunately.

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
Walfar: The UIDevice was used to, for instance, click on the map. This should launch a "onMapClick" method. But, since the map was never loaded by Cirrus, this method was not called. Instead, I directly call this method in my tests (with the condition that the map is not null), which makes the UIDevice useless, so I simply removed it.

Antho1426 (approved):

PR #241, opened 2021-05-13
Refactor/222 edit profile activity

kotleta2007:
This is the same pull request as the one that was opened for `feature/demo`.

The only difference is that I have added UserDatabase and its tests.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -100,6 +100,9 @@ dependencies {
     }
     annotationProcessor 'androidx.room:room-compiler:2.2.6'
     implementation 'de.hdodenhof:circleimageview:3.1.0'
+
+    // https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-databind
+    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.0.1'
 }
 
 tasks.withType(Test) {
--- a/app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
+++ b/app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
@@ -20,13 +20,24 @@ public class DistanceCalculator {
     private ArrayList<Pair> activityDistanceMap;
     private boolean sorted;
 
+    /**
+     * Called in MainMapFragment to display the closest activities to the user.
+     * @param userLatitude the latitude of the current location of the user
+     * @param userLongitude the longitude of the current location of the user
+     */
     public DistanceCalculator(Double userLatitude, Double userLongitude) {
         this.userLatitude = userLatitude;
         this.userLongitude = userLongitude;
         activityDistanceMap = new ArrayList<Pair>();
     }
 
+    /**
+     * Called in MainMapFragment to update the list of the closest activities.
+     * @param activities the list of all activities available to the user
+     */
     public void setActivities(ArrayList<Activity> activities) {
+        activityDistanceMap = new ArrayList<Pair>();
+
         for (Activity activity : activities) {
             activityDistanceMap.add(new Pair(activity, 0.0));
         }
@@ -34,6 +45,9 @@ public class DistanceCalculator {
         sorted = false;
     }
 
+    /**
+     * Called in MainMapFragment to return all activities.
+     */
     public ArrayList<Activity> getAllActivities() {
         if (sorted) {
             ArrayList<Activity> activities = new ArrayList<Activity>();
@@ -47,18 +61,28 @@ public class DistanceCalculator {
         }
     }
 
-
+    /**
+     * Called in MainMapFragment to return the list of N closest activities.
+     * @param n the number of activities to return
+     */
     @RequiresApi(api = Build.VERSION_CODES.N)
     public ArrayList<Activity> getTopActivities(Integer n) {
         if (sorted) {
-            Stream<Activity> topActivitiesStream = activityDistanceMap.subList(0, n).stream().map(Pair::getKey);
-            return topActivitiesStream.collect(Collectors.toCollection(ArrayList::new));
-
+            if (n >= activityDistanceMap.size()) {
+                return getAllActivities();
+            } else {
+                Stream<Activity> topActivitiesStream = activityDistanceMap.subList(0, n).stream().map(Pair::getKey);
+                return topActivitiesStream.collect(Collectors.toCollection(ArrayList::new));
+            }
         } else {
             return null;
         }
     }
 
+    /**
+     * Called in MainMapFragment to return the list of activities that fall in a given radius.
+     * @param radius the radius (in km) around the current location of the user
+     */
     public ArrayList<Activity> getActivitiesInRadius(Double radius) {
         if (sorted) {
             ArrayList<Activity> topActivities = new ArrayList<Activity>();
@@ -75,7 +99,9 @@ public class DistanceCalculator {
         }
     }
 
-
+    /**
+     * Called in MainMapFragment to compute the distances between the user and the activities.
+     */
     public void calculateDistances() {
         for (Pair pair : activityDistanceMap) {
             pair.setValue(calculateDistance(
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
```diff
@@ -20,20 +20,34 @@
     private ArrayList<Pair> activityDistanceMap;
     private boolean sorted;
 
+    /**
```
RoxaneBurri, line 23: Good documentation !

app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
```diff
@@ -47,18 +61,28 @@ public void setActivities(ArrayList<Activity> activities) {
...
 
-
+    /**
+     * Called in MainMapFragment to return the list of N closest activities.
+     * @param n the number of activities to return
+     */
     @RequiresApi(api = Build.VERSION_CODES.N)
     public ArrayList<Activity> getTopActivities(Integer n) {
         if (sorted) {
-            Stream<Activity> topActivitiesStream = activityDistanceMap.subList(0, n).stream().map(Pair::getKey);
```
RoxaneBurri, line 54: why do you pass it in a loop ?

app/src/main/java/com/sdp/movemeet/utility/UserDatabase.java
```diff
@@ -0,0 +1,92 @@
+package com.sdp.movemeet.utility;
```
RoxaneBurri, line 1: You can add some documentation in this class

app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
kotleta2007: Thank you!

RoxaneBurri (approved):

app/src/main/java/com/sdp/movemeet/utility/UserDatabase.java
kotleta2007: Good point, I will work on it

app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
kotleta2007: `activityDistanceMap` is an `ArrayList` of `Pair`s (it's kind of like a `HashMap`).

When it's sorted, it looks like:
`{(ActivityA, 0.1), (ActivityB, 0.2), (ActivityC, 0.3)}`

So we want to return the first element ("the key") of every `Pair` in the `ArrayList`, that's why we call `.map(Pair::getKey)`.

Hope this helps!

app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
```diff
@@ -47,18 +61,28 @@ public void setActivities(ArrayList<Activity> activities) {
...
-            return topActivitiesStream.collect(Collectors.toCollection(ArrayList::new));
-
+            if (n >= activityDistanceMap.size()) {
+                return getAllActivities();
```
Walfar, line 72: You can directly use getActivities() from ActivitiesUpdater that returns the list of all activities. Alternatively, you could store the list of activities obtained from setActivities, and return it. No need to compute from the list of pairs in a for loop, as you will return the whole list anyway.

app/src/main/java/com/sdp/movemeet/utility/UserDatabase.java
```diff
@@ -0,0 +1,92 @@
...
+            e.printStackTrace();
+        }
+         */
+    }
+
+    public void addUser(String email, String password) throws IOException {
+        if (isInUsers(email)) {
+            return;
+        } else {
+            users.put(email, password);
```
Walfar, line 83: I think this should be encrypted in the future haha

app/src/test/java/com/sdp/movemeet/modelsTest/UserDatabaseTest.java
```diff
@@ -0,0 +1,45 @@
...
+        assertEquals(false, db.isInUsers("goodbye"));
+    }
+
+    @Test
+    public void SavingAndLoadingIsCorrect() throws IOException {
+        UserDatabase db = new UserDatabase();
+
+        db.addUser("hello", "world");
+
+        db.writeToFile();
```
Walfar, line 40: Given that nothing is tested here, I assume that this doesn't really work yet ? How do we know that the map of users written corresponds to the one read ?

app/src/main/java/com/sdp/movemeet/utility/UserDatabase.java
```diff
@@ -0,0 +1,92 @@
...
+            e.printStackTrace();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+         */
+    }
+
+    public void writeToFile() throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.writeValue(new File(filename), users);
```
Walfar, line 55: Isn't it possible to simply put the new user added, instead of having to write the whole map and clean everytime ?

app/build.gradle
```diff
@@ -100,6 +100,9 @@ dependencies {
     }
     annotationProcessor 'androidx.room:room-compiler:2.2.6'
     implementation 'de.hdodenhof:circleimageview:3.1.0'
+
+    // https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-databind
+    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.0.1'
```
Antho1426, line 105: What is this implementation for?

app/src/main/java/com/sdp/movemeet/utility/UserDatabase.java
kotleta2007: Sure! We can take inspiration from the Security section of SwEng that we have studied last year.

app/src/main/java/com/sdp/movemeet/view/profile/EditProfileActivity.java
```diff
@@ -60,15 +65,19 @@ protected void onCreate(Bundle savedInstanceState) {
 
         fAuth = FirebaseAuth.getInstance();
         fStore = FirebaseFirestore.getInstance();
-        user = fAuth.getCurrentUser();
-        if (user != null) {
-            userId = user.getUid();
+
+        userManager = new FirestoreUserManager(fStore, FirestoreUserManager.USERS_COLLECTION, new UserSerializer());
```
Antho1426, line 69: Where do you actually make use of this instance in your code (to implement the Firebase Firestore abstraction for the user?)

app/src/main/java/com/sdp/movemeet/utility/DistanceCalculator.java
kotleta2007: That's a good point! I will refactor it, because it can help us performance-wise (the map will load quicker).

app/src/test/java/com/sdp/movemeet/modelsTest/UserDatabaseTest.java
kotleta2007: It does work, but I haven't added it to the test suite. 
During the next sprint, I am going to enrich the test suite so that it covers more cases and tests this class thoroughly.

app/build.gradle
kotleta2007: It's a library that allows you to read/write Maps from/to JSON files. I use it in the UserDatabase class.

app/src/main/java/com/sdp/movemeet/utility/UserDatabase.java
kotleta2007: That's correct, thanks for the idea. I'll think about optimizing this.

For most use-cases, there is usually 1-2 users on one device, so for the time being, we won't gain much from optimizing this. However, it would be nice to improve this in the future.

app/src/main/java/com/sdp/movemeet/view/profile/EditProfileActivity.java
kotleta2007: For the time being, I refactored this code without making use of userManager.

For the next sprint, I'll ask Kepler and look at your code to see how we can abstract away the Firebase Firestore calls. I believe that there is more abstraction to be done here

Walfar (approved):

PR #249, opened 2021-05-14
Link GPS recording data to activity

bitshrine:


app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityTest.java
```diff
@@ -55,7 +55,7 @@
     private final static Date DUMMY_DATE = new Date(2021, 11, 10, 1, 10);
     private final static double DUMMY_DURATION = 10.4;
     private final static Sport DUMMY_SPORT = Sport.Running;
-    private final static String DUMMY_ADDRESS = "address";
+    private final static String DUMMY_ADDRESS = "dummy_address";
```
Antho1426, line 58: Okey @Walfar! I guess we finally found from where the `null` activity arises! I took a screenshot of its document on Firebase Firestore and the address set as "dummy_address" seems to appear solely in this ActivityDescriptionActivityTest! Moreover, I checked this morning and we had still no `null` activity in the database. The time at which @vvhaleshark did his last commit (i.e. about 3 hours ago) corresponds to the time at which the `null` activity appeared on our database!

By the way, why did you rename this address @vvhaleshark? Was it precisely to track the `null` activity by setting a distinguishable address name?

app/src/main/java/com/sdp/movemeet/backend/providers/BackendInstanceProvider.java
```diff
@@ -6,6 +6,11 @@
 import com.google.firebase.firestore.FirebaseFirestore;
 import com.google.firebase.storage.FirebaseStorage;
 
+/**
+ * Singleton class which provides a single backend storage service instance for use inside
+ * other classes. The currently available instances are FirebaseFirestore, FirebaseDatabase,
+ * and FirebaseStorage.
+ */
```
Antho1426, line 13: After changing the constructor of FirestoreManager to get rid of the useless calls to the actual Firebase services, I get an error here for the `firestore` variable... Here is the error:

`Default FirebaseApp is not initialized in this process null. Make sure to call FirebaseApp.initializeApp(Context) first.`

Have you already encountered this before?

app/src/main/java/com/sdp/movemeet/backend/serialization/ActivitySerializer.java
```diff
@@ -70,6 +77,11 @@ public Activity deserialize(Map<String, Object> data) {
                 ((Timestamp) data.get(CREATION_KEY)).toDate()
         );
 
+        Map<String, GPSPath> recordings = (Map<String, GPSPath>) data.getOrDefault(GPS_RECORDINGS_KEY, null);
```
Antho1426, line 80: Is the GPS path recorded under the form of a String?

app/src/main/java/com/sdp/movemeet/models/GPSPath.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    public long getTime() {
+        return this.time;
+    }
+
+    /**
+     * @return the total distance covered by this path, in meters
+     */
+    public float getDistance() {
+        if (this.distance < 0) this.distance = computeTotalDistance(this.path);
```
Antho1426, line 54: I don't get the purpose of checking if the distance is less than 0. Is it because it is initialised as "-1", but then it would mean that this `getDistance()` method can be accessed only once?

app/src/main/java/com/sdp/movemeet/models/GPSPath.java
```diff
@@ -0,0 +1,147 @@
...
+        return this.distance;
+    }
+
+    /**
+     * @return the average speed over this path, in km/h
+     */
+    public float getAverageSpeed() {
+        if (this.averageSpeed < 0) this.averageSpeed = computeAverageSpeed(getDistance(), this.time);
```
Antho1426, line 62: Same question here...

app/src/main/java/com/sdp/movemeet/models/GPSPath.java
```diff
@@ -0,0 +1,147 @@
...
+     * time.
+     * @param distance The distance travelled, in meters
+     * @param time The time it took to travel, in milliseconds
+     * @return the average speed (in km/h)
+     */
+    public static float computeAverageSpeed(float distance, long time) {
+        if (time <= 0) throw new IllegalArgumentException();
+        if (distance <= 0) return 0;
+
+        return (distance / 1000) / (time / 3600);
```
Antho1426, line 121: Speed is hence in [km/h], is it right?

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
```diff
@@ -52,6 +54,8 @@
 public class ActivityDescriptionActivity extends AppCompatActivity {
 
     private static final String TAG = "ActDescActivity";
+    public static final String DESCRIPTION_ACTIVITY_KEY = "activitykey";
```
Antho1426, line 57: Oh, this is a good practice to define the constant strings for retrieving values from intents in as static fields at the beginning of the class, I will apply this in ChatActivity and EditProfileActivity!

But why did you set it as public? Do you plan to use this String in another class?

And did you also modify this hard-coded String in MainMapFragment (i.e. the activity from which the intent containing the "activity" object comes) in the same way?

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityUnregister.java
```diff
@@ -46,7 +46,7 @@ protected void onCreate(Bundle savedInstanceState) {
         Intent intent = getIntent();
 
         if (intent != null) {
-            activity = (Activity) intent.getSerializableExtra("activity");
+            activity = (Activity) intent.getSerializableExtra(ActivityDescriptionActivity.DESCRIPTION_ACTIVITY_KEY);
```
Antho1426, line 49: Ok I guess I have my answer now ^^' you use it precisely here 👍

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
```diff
@@ -207,6 +213,7 @@ private void createAddressView() {
      * Syncing registered participant to Firebase Firestore (field array "participantId")
      */
     public void registerToActivity(View v) {
+        userId = fAuth.getCurrentUser().getUid();
```
Antho1426, line 216: Why do you redefine ("re-get") the `userId` here? We already get it in the `onCreate`, aren't we?

app/src/main/java/com/sdp/movemeet/view/map/GPSRecordingActivity.java
```diff
@@ -70,15 +94,21 @@ public void onLocationResult(@NonNull LocationResult locationResult) {
...
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_g_p_s_recording);
 
+        Intent intent = getIntent();
+        activity = (Activity) intent.getSerializableExtra(ActivityDescriptionActivity.RECORDING_EXTRA_NAME);
+
+        if (AuthenticationInstanceProvider.getAuthenticationInstance().getCurrentUser() == null || activity == null) {
+            intent = new Intent(GPSRecordingActivity.this, MainActivity.class);
```
Antho1426, line 107: If the `activity` is `null`, I agree that we have to redirect the user to the `MainActivity`, but if the "Firebase User" is `null`, shouldn't we rather redirect the user directly to the `LoginActivity`?

app/src/main/java/com/sdp/movemeet/view/map/GPSRecordingActivity.java
```diff
@@ -104,7 +134,30 @@ public void toggleRecording(View view) {
...
             path.clear();
+        } else {
+            time = System.currentTimeMillis() - time;
+        }
+    }
+
+    @Override
+    public void onBackPressed() {
+        if (!recording) {
+            Intent intent = new Intent(GPSRecordingActivity.this, ActivityDescriptionActivity.class);
```
Antho1426, line 147: This might interests @jlstoupy98 ! An implementation of the `onBackPressed` method redirecting the user to another activity (here ActivityDescriptionActivity) with some extra information in the intent!

app/src/main/java/com/sdp/movemeet/view/map/MainMapFragment.java
```diff
@@ -178,7 +178,7 @@ else if (!marker.getTag().equals("my position")) {
             } else {
                 intent = new Intent(supportMapFragment.getActivity(), ActivityDescriptionActivity.class);
             }
-            intent.putExtra("activity", act);
+            intent.putExtra(ActivityDescriptionActivity.DESCRIPTION_ACTIVITY_KEY, act);
```
Antho1426, line 181: Ok nice, you did in fact modified this hard-coded String in the MainMapFragment, well done 👍. Based on this, I also modified the intents in `ProfileActivity` and `EditProfileActivity` accordingly!

Antho1426 (approved):

app/src/androidTest/java/com/sdp/movemeet/view/home/HomeScreenActivityLoggedTest.java
```diff
@@ -72,10 +72,4 @@ public void noAccountHasCorrectIntentWhenLogged() throws InterruptedException {
         onView(withId(R.id.noAccountButton)).perform(click());
         intended(hasComponent(MainActivity.class.getName()));
     }
-
```
RoxaneBurri, line 75: So now the record button is in the activity description ?

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
```diff
@@ -35,6 +36,14 @@
     @Rule
     public ActivityScenarioRule<LoginActivity> LoginTestRule = new ActivityScenarioRule<>(LoginActivity.class);
 
+    @Before
+    public void signOutBeforeTest() {
```
RoxaneBurri, line 40: What is the goal of this test ?

app/src/androidTest/java/com/sdp/movemeet/view/map/GPSRecordingActivityTest.java
```diff
@@ -1,11 +1,13 @@
 package com.sdp.movemeet.view.map;
 
 import android.Manifest;
+import android.content.Intent;
```
RoxaneBurri, line 4: There are a lot of import, have you deleted the non-useful ones ?

app/src/androidTest/java/com/sdp/movemeet/view/map/GPSRecordingActivityTest.java
```diff
@@ -62,9 +85,10 @@
     @Rule
     public GrantPermissionRule CoarseLocationPermissionRule = GrantPermissionRule.grant(Manifest.permission.ACCESS_COARSE_LOCATION);
 
+    Activity activity;
 
-    @Rule
-    public ActivityScenarioRule<GPSRecordingActivity> testRule = new ActivityScenarioRule<>(GPSRecordingActivity.class);
+    /*@Rule
+    public ActivityScenarioRule<GPSRecordingActivity> testRule = new ActivityScenarioRule<>(GPSRecordingActivity.class);*/
```
RoxaneBurri, line 91: Why this rule is commented ?

RoxaneBurri (approved):

app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityTest.java
bitshrine: Yes, I changed it for debugging purposes

app/src/androidTest/java/com/sdp/movemeet/view/home/HomeScreenActivityLoggedTest.java
bitshrine: Yes, since recordings are always tied to an activity, there is no need for a record button on the home screen which could not save the data with an activity

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
bitshrine: It's not a test, it's simply a function that runs before each test and makes sure there is no active user before the tests are run. I was having issues with sign-ins in previous tests messing up these ones, so I added this here.

app/src/androidTest/java/com/sdp/movemeet/view/map/GPSRecordingActivityTest.java
bitshrine: Admittedly, I don't think I cleaned up the imports before committing. Thank you for pointing it out!

app/src/androidTest/java/com/sdp/movemeet/view/map/GPSRecordingActivityTest.java
bitshrine: Ideally, I would like to be able to launch the Scenario for this test class without having to set it up in a @Before function, but it requires passing an Activity as an extra and I'm unsure of how to go about that exactly

app/src/main/java/com/sdp/movemeet/backend/providers/BackendInstanceProvider.java
bitshrine: Yes, it's due to the test being run as a unit test (which does not have a Context), meaning it cannot access app-specific things such as Firebase

app/src/main/java/com/sdp/movemeet/backend/serialization/ActivitySerializer.java
bitshrine: No, it's recorded as a GPSPath. The String is the uid of the user to whom the recording belongs (since many users can record paths for the same activity)

app/src/main/java/com/sdp/movemeet/models/GPSPath.java
bitshrine: The method can be called more than once, but this makes sure that accessing it doesn't return an uninitialized value.

app/src/main/java/com/sdp/movemeet/models/GPSPath.java
bitshrine: Exactly!

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
bitshrine: Yes, it is also used in the GPSRecording class, when returning from a recording. This way, when recreating the ActivityDescriptionActivity, it can use the newly recorded data passed in the updated Activity.

Yes, all interactions which required passing an Activity to the ActivityDescriptionActivity should have been updated with this constant.

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
bitshrine: Indeed. This is probably duplicate code due to a merge with the main, which I didn't catch afterwards

app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityTest.java
Antho1426: Nice, we might have found our `null` activity then, thank you!

app/src/main/java/com/sdp/movemeet/backend/providers/BackendInstanceProvider.java
Antho1426: Yes thanks for your commit on my branch and for helping me fix this!

app/src/main/java/com/sdp/movemeet/backend/serialization/ActivitySerializer.java
Antho1426: Oh I see, all clear I understand the working of your `Map` now.

app/src/main/java/com/sdp/movemeet/models/GPSPath.java
Antho1426: Ok, I see, nice trick!

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
Antho1426: Ok alright, make sure to change this then (if you are able to) 😉.

app/src/androidTest/java/com/sdp/movemeet/view/home/HomeScreenActivityLoggedTest.java
Antho1426: Yes, and this is also important to fulfil the "sensor requirement" by fully integrating the GPS tracking feature into the core of our app (and not having it "simply" on the HomeScreen).

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
Antho1426: Yes, I see! @jlstoupy98, this might be probably a good idea to insert such kind of `signOutBeforeTest` at the beginning of the test that delete the user. This way, we might avoid our issue that inadvertently regularly erases the "official movemeet" account!

Antho1426 (approved):

PR #250, opened 2021-05-15
Try to improve coverage

jlstoupy98:


app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityTest.java
```diff
@@ -149,8 +149,8 @@ public void chatButtonIsCorrect() {
             assert (false);
         }
 
-        onView(withId(R.id.activityChatDescription)).perform(scrollTo(), click());
-        intended(hasComponent(ChatActivity.class.getName()));
+        //onView(withId(R.id.activityChatDescription)).perform(scrollTo(), click());
+        //intended(hasComponent(ChatActivity.class.getName()));
```
Antho1426, line 153: You did well to comment this test because we effectively observed that accessing the chat makes our "`java.lang.RuntimeException: Uncaught exception in Firebase Database runloop (19.7.0). If you are not already on the latest version of the Firebase SDKs, try updating your dependencies`" error reappear...

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
```diff
@@ -70,7 +70,6 @@
     }
 
     @Test
-    @LargeTest
```
Antho1426, line 73: Do you know what it could change to use the @Test annotation instead of the @LargeTest annotation ?

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
```diff
@@ -53,12 +67,16 @@
...
-    public void truc(){}
-    /*@Before
+
+    @Before
     public void signIn(){
+        Navigation.profileField = false;
         MainActivity.enableNav = true;
         UploadActivityActivity.enableNav = true;
+        ActivityDescriptionActivity.enableNav = true;
+        ChatActivity.enableNav = true;
```
Antho1426, line 77: Do you really want to enable the navigation to the chat knowing that, as we observed so far, any navigation to the chat may induce a "`java.lang.RuntimeException: Uncaught exception in Firebase Database runloop (19.7.0). If you are not already on the latest version of the Firebase SDKs, try updating your dependencies`" error? Is it required?

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
```diff
@@ -125,10 +147,74 @@ public void NavigationTest() {
...
+
+        // Open Drawer to click on navigation.
+        onView(withId(R.id.drawer_layout)).check(matches(isClosed(Gravity.LEFT))).perform(DrawerActions.open());
+
+        try{
+            Thread.sleep(500);
+        }catch(Exception e){}
+
+        onView(withId(R.id.nav_view)).perform(swipeUp());
+        onView(withId(R.id.nav_chat)).perform(click());
```
Antho1426, line 174: Did you observe the apparition of the "`java.lang.RuntimeException: Uncaught exception in Firebase Database runloop (19.7.0). If you are not already on the latest version of the Firebase SDKs, try updating your dependencies`" error with this line of code? Do you think navigating to the chat here could induce this error?

app/src/androidTest/java/com/sdp/movemeet/view/navigation/security/ChatActivitySecurityTest.java
```diff
@@ -23,16 +27,20 @@
...
         if (firebaseAuth.getCurrentUser() != null) firebaseAuth.signOut();
     }
 
-    //@Rule
-    //public ActivityScenarioRule<ChatActivity> testRule = new ActivityScenarioRule<>(ChatActivity.class);
-
     @Test
     public void redirectionTest() {
-        //onView(withId(R.id.button_login)).check(matches(isDisplayed()));
+        ChatActivity.enableNav = false;
```
Antho1426, line 36: Could you remind me what is the purpose of setting `ChatActivity.enableNav` to false here? I'm actually not really sure to get the purpose of the `.enableNav` attribute...

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
```diff
@@ -112,6 +111,8 @@ protected void onCreate(Bundle savedInstanceState) {
         if (fAuth.getCurrentUser() == null) {
             startActivity(new Intent(getApplicationContext(), LoginActivity.class));
             finish();
+        } else {
+            userId = fAuth.getCurrentUser().getUid();
```
Antho1426, line 115: Nice refactor yes! The code is more logical and cleaner this way!

app/src/main/java/com/sdp/movemeet/view/chat/ChatActivity.java
```diff
@@ -123,7 +123,11 @@ protected void onCreate(Bundle savedInstanceState) {
         messageManager = new FirebaseDBMessageManager(database, new MessageSerializer());
 
         fAuth = FirebaseAuth.getInstance();
-        if (fAuth.getCurrentUser() != null) {
+        // The aim is to block any direct access to this page if the user is not logged in
+        if (fAuth.getCurrentUser() == null) {
+            startActivity(new Intent(getApplicationContext(), LoginActivity.class)); // sending the user to the "Login" activity
+            finish();
+        } else {
```
Antho1426, line 130: Again, that's a good way to handle the Authentication! Well done!

app/src/main/java/com/sdp/movemeet/view/navigation/Navigation.java
```diff
@@ -182,7 +184,7 @@ public void createDrawer() {
         navigationView.setNavigationItemSelectedListener(this::onNavigationItemSelected);
         navigationView.setCheckedItem(this.activityId);
 
-        fillNavigationProfileFields(new TextView[] {fullName, email, phone});
+        if (profileField) fillNavigationProfileFields(new TextView[] {fullName, email, phone});
```
Antho1426, line 187: In which cases could the `profileField` variable suddenly be set to false?

app/src/main/java/com/sdp/movemeet/view/profile/ProfileActivity.java
```diff
@@ -88,7 +88,11 @@ protected void onCreate(Bundle savedInstanceState) {
         userManager = new FirestoreUserManager(fStore, FirestoreUserManager.USERS_COLLECTION, new UserSerializer());
 
         fAuth = FirebaseAuth.getInstance();
-        if (fAuth.getCurrentUser() != null) {
+        //The aim is to block any direct access to this page if the user is not logged
+        if (fAuth.getCurrentUser() == null) {
+            startActivity(new Intent(getApplicationContext(), LoginActivity.class));
+            finish();
+        } else {
```
Antho1426, line 95: Again, good refactor, love it! 👍

Antho1426 (approved):

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
jlstoupy98: The first use was to increase the time limit of the test, the "normal" test has a smaller time limit than the large test. If I remember correctly, these annotations can be "small", "medium" or "large" and this can also have an impact on some network access. For example, I believe that a "small test" does not allow access to databases while a medium or large one does.

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
jlstoupy98: In any case, it is not this line that could create the error. It just allows you to specify that at the instantiation of the chat activity, the menu will be created. So in order to have a menu from the chat, this line is necessary.

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
jlstoupy98: It is true that it seems that this error comes from the navigation to the chat. But, maybe by chance, I have never had a problem with this test yet. Fingers crossed that it lasts.

app/src/androidTest/java/com/sdp/movemeet/view/navigation/security/ChatActivitySecurityTest.java
jlstoupy98: This technique was devised by Kepler. Each class has an enablenav attribute that allows to decide if the navigation menu should be instantiated or not. Indeed, for example in this case, the purpose of the test is to verify that a user who is not logged in and who arrives on the chat page is redirected to the login. In this case, when the menu is instantiated, the header will try to retrieve the information of a user that does not exist, leading to an error before the test even starts. By doing this, the menu will not be created so the navigation will not try to retrieve the data of a null user allowing the test to run properly

app/src/main/java/com/sdp/movemeet/view/chat/ChatActivity.java
jlstoupy98: Haha thank you ! In addition, in one of the classes, we tried to retrieve the id of the user before even testing if the authentication instance existed which resulted in an error.

app/src/main/java/com/sdp/movemeet/view/navigation/Navigation.java
jlstoupy98: It is a variable identical to enablenav. It can be set to false in tests for example if we don't want to try to get the user's information when creating the menu.

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
Antho1426: Ok ok, thank you for those details, I wasn't sure!

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
Antho1426: Ok ok, effectively 🤞

app/src/androidTest/java/com/sdp/movemeet/view/navigation/security/ChatActivitySecurityTest.java
Antho1426: Awesome, I see! Thank you for this thorough explanation!

app/src/main/java/com/sdp/movemeet/view/chat/ChatActivity.java
Antho1426: Oh yes, good catch, thank you for correcting the!

app/src/main/java/com/sdp/movemeet/view/navigation/Navigation.java
Antho1426: Ok ok I see, good trick!

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
```diff
@@ -112,6 +133,7 @@ public void NavigationTest() {
         try{
             Thread.sleep(500);
         }catch(Exception e){}
+        intended(hasComponent(UploadActivityActivity.class.getName()));
```
Walfar, line 136: I wonder why you don't have the null instance error, when loading the navigation with a null user. That was a big issue for me when I tried making the navigation tests

app/src/androidTest/java/com/sdp/movemeet/view/navigation/security/ActivityDescriptionSecurityTest.java
```diff
@@ -20,16 +26,20 @@
...
         if (firebaseAuth.getCurrentUser() != null) firebaseAuth.signOut();
     }
 
-    //@Rule
-    //public ActivityScenarioRule<ActivityDescriptionActivity> testRule = new ActivityScenarioRule<>(ActivityDescriptionActivity.class);
-
     @Test
     public void redirectionTest() {
-        //onView(withId(R.id.button_login)).check(matches(isDisplayed()));
+        ActivityDescriptionActivity.enableNav = false;
```
Walfar, line 35: Why is it better to put this assignment directly inside the test, rather than in the @Before ? Is it for consistency with the name of the method ("checkUserNotLogged"), or is there another reason ?

app/src/main/java/com/sdp/movemeet/view/main/MainActivity.java
```diff
@@ -29,29 +29,21 @@
...
-        if (enableNav) new Navigation(this, R.id.nav_home).createDrawer();
-
         //The aim is to block any direct access to this page if the user is not logged
         fAuth = FirebaseAuth.getInstance();
         if (fAuth.getCurrentUser() == null) {
             startActivity(new Intent(getApplicationContext(), LoginActivity.class)); // sending the user to the "Login" activity
             finish();
         }
 
+        if (enableNav) new Navigation(this, R.id.nav_home).createDrawer();
```
Walfar, line 46: Does it change something to have the creation of the Navigation drawer after the user check ?

Walfar (approved):
Everything seems to work well :) Good job improving the coverage ! I struggled to make those tests work, but you did fine apparently. My PR will then be useless, I will make another one only for the message serializer.

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
jlstoupy98: As discussed on Monday, I used the same technique as for the navigation. Basically, there is a variable to choose if the header will be displayed or not (and thus if the user's info should be retrieved). It may be a bit dirty but it worked for me even if I should look more closely because some behavior seems strange.

app/src/androidTest/java/com/sdp/movemeet/view/navigation/security/ActivityDescriptionSecurityTest.java
jlstoupy98: I don't think it's really better or worse. At first I had it in the test just by inattention in my tests but after, I indeed found that compared to the name of the method, it was maybe cleaner to keep this assignment directly in the test.

app/src/main/java/com/sdp/movemeet/view/main/MainActivity.java
jlstoupy98: Yes, it really changes because the navigation will try to retrieve information from a user. But if this one is null, there will be an error when creating the navigation. By reversing the order, we check that a user is connected and therefore that we have an authentication instance to create the navigation using this instance.

PR #251, opened 2021-05-15
246 - Firebase storage abstraction implementation 

### LINKED ISSUE #246: Create Firebase Storage abstraction for local cache

### LINKED ISSUE #248: Correct header picture loading in ActivityDescriptionActivity

### LINKED ISSUE #255: Implement Firebase Firestore abstraction in EditProfileActivity

### LINKED ISSUE #261: Implement assertion for testing the upload of images (in EditProfileImagePickerTest)

Antho1426:
Implementation of the abstraction of Firebase Storage to be used jointly with our local cache feature (locally saving the header pictures of the sports activities).


app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityNotNullDescriptionTest.java
```diff
@@ -41,7 +42,7 @@
     private final static double DUMMY_DURATION = 10.4;
     private final static Sport DUMMY_SPORT = Sport.Running;
     private final static String DUMMY_ADDRESS = "address";
-    public FirebaseAuth mAuth;
+    public FirebaseAuth fAuth;
```
RoxaneBurri, line 45: Nice rename !

app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityTest.java
```diff
@@ -87,7 +88,7 @@
     public void signIn() {
         CountDownLatch latch = new CountDownLatch(1);
 
-        fAuth = FirebaseAuth.getInstance();
+        fAuth = AuthenticationInstanceProvider.getAuthenticationInstance();
```
RoxaneBurri, line 86: You make an abstraction for the sign in part ?

app/src/main/java/com/sdp/movemeet/backend/FirebaseInteraction.java
```diff
@@ -41,58 +41,58 @@
...
-     * @param profileDescription TextView corresponding to the profile description set by the user
-     * @return A HashMap containing the updated user information
-     */
-    public static Map<String, Object> updateDataInFirebase(TextView profileFullName, TextView profileEmail, TextView profilePhone, TextView profileDescription) {
-        Map<String, Object> edited = new HashMap<>();
-        edited.put("fullName", profileFullName.getText().toString());
-        edited.put("email", profileEmail.getText().toString());
-        edited.put("phone", profilePhone.getText().toString());
-        edited.put("description", profileDescription.getText().toString());
+//    /**
```
RoxaneBurri: Why do you comment the firebase interaction ?

app/src/main/java/com/sdp/movemeet/models/Image.java
```diff
@@ -0,0 +1,71 @@
+package com.sdp.movemeet.models;
```
RoxaneBurri, line 1: Nice, we have a class for the image, it would be easier to use it now !

app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityNotNullDescriptionTest.java
Antho1426: Yes! This is again to follow our agreed convention!

app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityTest.java
Antho1426: No, that's not an abstraction! This is simply a call to a method created by Kepler to instantiate the Firebase Authentication service in a cleaner way!

app/src/main/java/com/sdp/movemeet/backend/FirebaseInteraction.java
Antho1426: This was in case something went wrong when I implemented the Firebase Storage abstraction (so that I could quickly go back to what we had previously in FirebaseInteraction, to refer to it). But, now, I was able to completely remove FirebaseInteraction! So the question is not even worth asking! 😂

app/src/main/java/com/sdp/movemeet/models/Image.java
Antho1426: Yes that's it! A new model class to work more easily with the Firebase Storage abstraction!

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
```diff
@@ -25,12 +25,12 @@
...
 
-    public static final String mEmail = "email";
-    public static final String mPassword = "password";
+    public static final String falseEmail = "email";
+    public static final String falsePassword = "password";
     public static final String shortPassword = "pass";
 
-    public static final String Email = "movemeet@gmail.com";
-    public static final String Password = "password";
+    public static final String trueEmail = "movemeet@gmail.com";
```
bitshrine, line 32: Even though we have no abstraction for an authentication service, is there no way to use mocks in these tests to avoid having to actually sign in and out?

app/src/androidTest/java/com/sdp/movemeet/view/profile/DrawableMatcher.java
```diff
@@ -0,0 +1,68 @@
...
+import android.graphics.Bitmap;
+import android.graphics.Canvas;
+import android.graphics.drawable.Drawable;
+import android.view.View;
+import android.widget.ImageView;
+
+import org.hamcrest.Description;
+import org.hamcrest.TypeSafeMatcher;
+
+public class DrawableMatcher extends TypeSafeMatcher<View> {
```
bitshrine, line 20: Would you mind documenting this class?

app/src/main/java/com/sdp/movemeet/view/profile/EditProfileActivity.java
```diff
@@ -151,15 +149,19 @@ public void onFailure(@NonNull Exception e) {
 
     @VisibleForTesting(otherwise=VisibleForTesting.PRIVATE) // making this method always public for testing and private otherwise
     public void accessFirestoreUsersCollectionForUpdate() {
-        DocumentReference docRef = fStore.collection("users").document(userId);
-        Map<String, Object> edited = FirebaseInteraction.updateDataInFirebase(profileFullName, profileEmail, profilePhone, profileDescription);
-        docRef.update(edited).addOnSuccessListener(new OnSuccessListener<Void>() {
+        User user = new User(profileFullName.getText().toString(), profileEmail.getText().toString(), profilePhone.getText().toString(), profileDescription.getText().toString());
+        userManager.add(user, FirestoreUserManager.USERS_COLLECTION + "/" + userId).addOnSuccessListener(new OnSuccessListener() {
```
bitshrine, line 153: If all of the calls to `userManager.add()` look like this,, you can probably re-write the UserManager's implementation of the add method so that it automatically formats the path with the `collection + '/' + id` format

app/src/androidTest/java/com/sdp/movemeet/view/home/LoginActivityTest.java
Antho1426: It could be possible in fact to mock the authentication! As you showed us this evening, it is possible to mock almost any class using mockito. I have to investigate this yes!

app/src/androidTest/java/com/sdp/movemeet/view/profile/DrawableMatcher.java
Antho1426: Yes probably! This is a class I added to be able to assert that the picked image corresponds to the uploaded image in EditProfileActivityImagePickerTest. Initially I was not sure this mechanism would work and since I implemented it tonight (I have found it on the internet), I haven't taken the time to document it yet! But I will definitely do it!

app/src/main/java/com/sdp/movemeet/view/profile/EditProfileActivity.java
Antho1426: Yes, might be cleaner indeed! That's a very good remark, I will see what I can do!

RoxaneBurri (approved):

app/src/main/java/com/sdp/movemeet/backend/BackendManager.java
```diff
@@ -23,7 +23,16 @@
...
+    Task<?> set(T object, String path);
+
+    /**
+     * Update (i.e. add) a String value to a Firebase Firestore array field
+     * @param value the String value to update in the array.
+     * @param path the path of the instance in the backend.
+     * @param field the field to which the array corresponds in the backend.
+     * @return a Task, the success of which determines the success of the operation.
+     */
+    Task<?> updt(String value, String path, String field);
```
bitshrine: Why not call it `update`?

bitshrine: Also, I would switch the parameter order to path - field - value, but that's just my personal preference

app/src/androidTest/java/com/sdp/movemeet/view/profile/DrawableMatcher.java
```diff
@@ -0,0 +1,68 @@
+package com.sdp.movemeet.view.profile;
```
Walfar, line 1: So if I understand correctly, this is a utility class only used for comparing the images in the newly added tests, right ?

app/src/main/java/com/sdp/movemeet/backend/firebase/storage/StorageImageManager.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public StorageTask add(Image image, String path) {
+        return storageReference.child(path).putFile(image.getImageUri());
+    }
+
+    @Override
+    public Task<?> set(Image object, String path) {
+        return null;
```
bitshrine: If this operation is unsupported, maybe you can throw a `UnsupportedOperationException`?

app/src/androidTest/java/com/sdp/movemeet/view/profile/EditProfileActivityImagePickerTest.java
```diff
@@ -21,46 +30,83 @@
...
+            }
+        });
+
+        try {
+            latch.await();
+        } catch (InterruptedException e) {
+            assert(false);
+        }
+
+        ActivityScenario scenario = ActivityScenario.launch(EditProfileActivity.class);
```
Walfar, line 77: Shouldn't this be inside the OnSuccess ? As you can only go to this activity once you are signed in

app/src/androidTest/java/com/sdp/movemeet/view/profile/EditProfileActivityImagePickerTest.java
```diff
@@ -21,46 +30,83 @@
...
+
     }
 
     @Test
     public void uploadImageToFirebaseStorage() {
         File dir = activityRule.getActivity().getExternalCacheDir();
-        File file = new File(dir.getPath(), "pickImageResult.jpeg");
+        File file = new File(dir.getPath(), "pickImageResult.png");
```
Walfar, line 84: The conversion for other formats than .png is not possible ?

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.android.gms.tasks.OnFailureListener;
+import com.google.android.gms.tasks.OnSuccessListener;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.storage.UploadTask;
+import com.sdp.movemeet.backend.BackendManager;
+import com.sdp.movemeet.backend.firebase.storage.StorageImageManager;
+import com.sdp.movemeet.models.Image;
+import com.squareup.picasso.Picasso;
+
+public class ImageHandler {
```
bitshrine, line 24: Great job! 🥇

(Could use a little documentation though)

app/src/main/java/com/sdp/movemeet/backend/firebase/firestore/FirestoreManager.java
```diff
@@ -49,22 +50,33 @@ public FirestoreManager(FirebaseFirestore db, String collection, BackendSerializ
...
         if (object == null) throw new IllegalArgumentException();
-        // TODO: check why the number of registered participants vary on Firebase Firestore
         Map<String, Object> data = serializer.serialize(object);
+
         return db.document(path).set(data);
     }
 
+    @Override
+    public Task<Void> updt(String value, String path, String field) {
+        if (value == null || path == null) throw new IllegalArgumentException();
```
Walfar: Can't the field be null ?

app/src/main/java/com/sdp/movemeet/backend/firebase/storage/StorageImageManager.java
```diff
@@ -0,0 +1,49 @@
...
+        this.storageReference = BackendInstanceProvider.getStorageInstance().getReference();
+    }
+
+    @Override
+    public StorageTask add(Image image, String path) {
+        return storageReference.child(path).putFile(image.getImageUri());
+    }
+
+    @Override
+    public Task<?> set(Image object, String path) {
```
Walfar, line 26: I think it could be a good idea to implement those methods as well, although there are not used (yet). Otherwise, you can change "return null" by "throw new UnsupportedOperationException()".

app/src/main/java/com/sdp/movemeet/models/Image.java
```diff
@@ -0,0 +1,75 @@
...
+    private String documentPath;
+
+    /**
+     * Constructor for a new image
+     *
+     * @param imageUri URI pointing to the image file
+     */
+    public Image(Uri imageUri, ImageView imageImageView) {
+
+        if (imageUri == null && imageImageView == null) {
```
Walfar, line 23: Should be || instead of &&

app/src/main/java/com/sdp/movemeet/models/Image.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param imageUri set the URI of the image
+     */
+    public void setImageUri(Uri imageUri) {
+        this.imageUri = imageUri;
+    }
+
+    /**
+     * @param imageImageView set the URI of the image
+     */
+    public void setImageImageView(ImageView imageImageView) {
```
Walfar, line 62: Isn't this method name a bit redundant ? I think, as it's a method directly called for Image, it's okay to name it setImageView (same goes for setImageUri, becoming setUri etc...)

app/src/main/java/com/sdp/movemeet/backend/BackendManager.java
Antho1426: That's a good point, I will modify it straight away!

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private static final String TAG = "FirebaseInteraction";
+
+    private static BackendManager<Image> imageBackendManager;
```
Walfar, line 28: I wonder if we should define this as BackendManager<Image>, or directly as StorageImageManager ? I guess the first option is best in term of abstraction, but the second prevent useless casts.

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+            progressBar.setVisibility(View.VISIBLE);
+        }
+        imageBackendManager = new StorageImageManager();
+        Task<Uri> document = (Task<Uri>) imageBackendManager.get(image.getDocumentPath());
+        document.addOnSuccessListener(new OnSuccessListener<Uri>() {
+            @Override
+            public void onSuccess(Uri uri) {
+                Log.d(TAG, "Image successfully fetched from Firebase Storage!");
+                image.setImageUri(uri);
+                Picasso.get().load(uri).into(image.getImageView());
```
Walfar, line 47: I wil have to change the dependency of Picasso to Glide, as it's the only way for me to make it work the cache.

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import com.sdp.movemeet.models.Image;
+import com.squareup.picasso.Picasso;
+
+public class ImageHandler {
+
+    private static final String TAG = "FirebaseInteraction";
+
+    private static BackendManager<Image> imageBackendManager;
+
+    public static void loadImage(Image image, ProgressBar progressBar) {
```
Walfar, line 36: Might have to think about the effectiveness of the Image class, and the fact that only having the direct call to ImageView, ImagePath and ImageUri might be enough.

app/src/main/java/com/sdp/movemeet/backend/firebase/storage/StorageImageManager.java
Antho1426: Good point! I just modified it! Thanks Kepler, that's really thanks to your remark that I learn the right and proper ways to code in Java!

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Antho1426: Yes! Thanks Kepler! I documented! If you @Walfar could have a look at this documentation (or eventually even adjust it later on according to your modifications concerning the local cache, it would be helpful!)

app/src/androidTest/java/com/sdp/movemeet/view/profile/DrawableMatcher.java
Antho1426: That's it Victor! And this class is actually used only for the test `EditProfileActivityImagePickerTest`. It actually implements an Espresso Matcher for ImageView and I found it [here](https://medium.com/@dbottillo/android-ui-test-espresso-matcher-for-imageview-1a28c832626f). I will document it today!

app/src/androidTest/java/com/sdp/movemeet/view/profile/EditProfileActivityImagePickerTest.java
Antho1426: Well, you might be right, but I tried to modify the test in this way and, in fact, (even though it would make more sense to me too as well) it doesn't pass anymore in case I put this line inside the `OnSuccess`... I think I will leave this test as it is for the moment and will try to come back to this later on if I manage to implement it in another way 👍.

app/src/androidTest/java/com/sdp/movemeet/view/profile/EditProfileActivityImagePickerTest.java
Antho1426: I guess it is, but I noticed that it works well with `.png` images and I stick to it! In fact, it matches the method I used below `bm.compress(Bitmap.CompressFormat.PNG, 100, outStream);` and the image I choose to set is a .png picture: icon_tricking.png.

app/src/main/java/com/sdp/movemeet/backend/firebase/firestore/FirestoreManager.java
Antho1426: You're right, it shouldn't be `null`, I updated this method accordingly this way:

```java
@Override
    public Task<Void> update(String path, String field, String value) {
        if (path == null || field == null || value == null) throw new IllegalArgumentException();

        return db.document(path).update(field, FieldValue.arrayUnion(value));
    }
```

app/src/main/java/com/sdp/movemeet/backend/firebase/storage/StorageImageManager.java
Antho1426: You're right Victor! Kepler made the same comment! For the moment I modified this file with some `throw new UnsupportedOperationException()` for the methods we don't use! I guess we don't need those methods for the Image objects for the moment, but if we need them in the future, we will implement them and remove the `throw new UnsupportedOperationException()` parts 👍.

app/src/main/java/com/sdp/movemeet/models/Image.java
Antho1426: No, this is in fact intended. In some situation, the `imageUri` can be `null` (for cases in which we **fetch** the image from Firebase Storage, such as when we retrieve the user profile picture in `ProfileActivity` and `EditProfileActivity`, or the header picture in `ActivityDescriptionActivity` or the chat image in `MessageViewHolder` from Firebase Storage, because we just have the path of the image in Firebase Storage, but not yet its URI) and in some other situations, the `imageImageView` can be `null` as well (for cases in which we **upload** the image to Firebase Storage, such as in `ChatActivity` for instance where we specify the URI of the image to upload to Firebase Storage, but not its ImageView (which is useless at this point)), but they can't be both simultaneously `null`! That's the reason why I use here a `&&` and not a `||`symbol.

app/src/main/java/com/sdp/movemeet/models/Image.java
Antho1426: You're completely right, but I took the decision to name it this way to stay consistent with the setters name we chose so far. For instance for the Message class, we have `setMessageUser`, `setMessageText` and `setMessageUserId`. In the case of the Image class, we have also `setImageUri`. It seems like the convention we chose works this way:

`set` + `<class-name>` + `<field-type>`

That's the reason why in this case I chose to set the name of this setter as `setImageImageView` even if this is a bit redundant effectively!

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Antho1426: Well, I'm not sure as well, but this is to stay consistent with the way we implemented all our different kinds of managers so far (such as: `userManager`, `activityBackendManager`, `messageManager`, etc.). I think Kepler would have done it this way as well! This way of defining the different backend managers really highlights our abstraction I guess.

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Antho1426: Ok ok, no problem, if you manage to make `ImageHandler` still works with Glide then! Personally I use Glide only for the temporary spinning wheel when we are sending an image via the chat.

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Antho1426: Yes, you're right... As we discussed during our last standup meeting, it might not be worth keeping this `Image` class in the future if we only make direct calls to imageImageView, imagePath and imageUri. Let me know how your cache implementation is evolving and, if we have time and if it makes sense, we can simply remove this `Image` class!

Walfar (approved):
Looks all good to me :)

bitshrine (approved):

PR #252, opened 2021-05-19
List of activities

RoxaneBurri:


app/src/androidTest/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivityTest.java
```diff
@@ -37,8 +31,6 @@
 import static androidx.test.espresso.action.ViewActions.click;
 import static androidx.test.espresso.action.ViewActions.scrollTo;
 import static androidx.test.espresso.assertion.ViewAssertions.matches;
-import static androidx.test.espresso.intent.Intents.intended;
-import static androidx.test.espresso.intent.matcher.IntentMatchers.hasComponent;
 import static androidx.test.espresso.matcher.ViewMatchers.withId;
 import static androidx.test.espresso.matcher.ViewMatchers.withText;
```
Antho1426: Optimizing Imports is important if fact! We should do it systematically after editing a file, this will keep our work cleaner and more organised!

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -164,7 +164,7 @@ public void cameraZoomsOnlyOnFirstCall() {
     public void clickingOnActivityMakesCorrectIntentWhenLogged() {
         MainMapFragment mapFragment = fragmentTestRule.getFragment();
         Activity act = new Activity("activity id", "organizer id", "my title", 4, new ArrayList<>(), 0, 0,
-                    "desc", null, new Date(), 1, Soccer, "Dubai UAE", new Date());
+                    "desc", "activities/123455", new Date(), 1, Soccer, "Dubai UAE", new Date());
```
Antho1426, line 168: As discussed with Victor finally, even if this won't change anything in the end (because this activity object here is not uploaded to Firebase Firestore), it's a very good practice to have changed the documentPath like this! In fact, the documentPath of activities in Firebase Firestore are always of the form "activities/<some-string>".

app/src/main/java/com/sdp/movemeet/models/User.java
```diff
@@ -17,40 +16,43 @@
...
-     * @param phone : phone number of the user
+     *
+     * @param fullName    : full name of the user
+     * @param email       : google email of the user
+     * @param phone       : phone number of the user
      * @param description : description of the user
      */
-    public User (String fullName, String email, String phone, String description) {
-        if(fullName == null || email == null || phone == null){
+    public User(String fullName, String email, String phone, String description, ArrayList<String> registeredActivities) {
```
Antho1426: Yes! As discussed, I think it's a good idea to add this field to User objects. This will allow easy access to the list of activities the user is registered for! (Rather than downloading all the activities and then looking for each one to see if the user is registered for it! (As you might have want to implement this at the beginning)) 👍

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
```diff
@@ -60,15 +55,10 @@
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     public static boolean enableNav = true;
 
-    DrawerLayout drawerLayout;
-    NavigationView navigationView;
-    Toolbar toolbar;
-    TextView textView;
-
-    TextView fullName, email, phone, organizerView, numberParticipantsView, participantNamesView;
+    TextView organizerView, numberParticipantsView, participantNamesView;
```
Antho1426: Do you think we should put each of those TextViews on a different line? Would it be cleaner?

app/src/main/java/com/sdp/movemeet/backend/serialization/UserSerializer.java
```diff
@@ -15,18 +16,22 @@
     public static final String PHONE_KEY = "phone";
     // The key used to access the description attribute of a serialized User
     public static final String DESCRIPTION_KEY = "description";
+    // The key used to access the list of registered activities attribute of a serialized User
+    public static final String REGISTERED_ACTIVITIES_KEY = "activityId";
```
Antho1426: Wouldn't it be more logical to name this key "activitiesId" (and not "activityId") since the user is certainly going to be registered to many activities?

Antho1426 (approved):

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
RoxaneBurri: Yes !

app/src/main/java/com/sdp/movemeet/models/User.java
RoxaneBurri: Yes it is easier and faster by this method ! Thanks for the advice !

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
RoxaneBurri: I do it ! Thanks !

app/src/main/java/com/sdp/movemeet/backend/serialization/UserSerializer.java
RoxaneBurri: Yes !

PR #254, opened 2021-05-19
Fix/refactor navigation flag

bitshrine:
Instead of using `enableNav` on a per-class basis to enable/disable navigation during tests, we use a single boolean flag `profileField` in the `Navigation` activity.

### DIFF
--- a/app/src/androidTest/java/com/sdp/movemeet/view/activity/UploadActivityActivityTest.java
+++ b/app/src/androidTest/java/com/sdp/movemeet/view/activity/UploadActivityActivityTest.java
@@ -32,6 +32,7 @@ import com.sdp.movemeet.backend.serialization.BackendSerializer;
 import com.sdp.movemeet.models.Activity;
 import com.sdp.movemeet.view.activity.UploadActivityActivity;
 import com.sdp.movemeet.view.main.MainActivity;
+import com.sdp.movemeet.view.navigation.Navigation;
 import com.sdp.movemeet.view.navigation.security.MainActivitySecurityTest;
 
 import org.hamcrest.Matcher;
@@ -89,8 +90,7 @@ public class UploadActivityActivityTest {
     @Before
     public void setup() {
         // Disable navigation for tests
-        UploadActivityActivity.enableNav = false;
-        MainActivity.enableNav = false;
+        Navigation.profileField = false;
 
         // Set up mocks and their behaviors
         // Set up fake database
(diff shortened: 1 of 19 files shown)


app/src/androidTest/java/com/sdp/movemeet/view/activity/UploadActivityActivityTest.java
```diff
@@ -89,8 +90,7 @@
     @Before
     public void setup() {
         // Disable navigation for tests
-        UploadActivityActivity.enableNav = false;
-        MainActivity.enableNav = false;
+        Navigation.profileField = false;
```
Antho1426, line 93: So, as discussed yesterday evening, the purpose of this Navigation.profileField Boolean is to be able to navigate freely inside the app without having to consider a specific user and hence to have to print his information in the side panel (which makes it possible to avoid that tests do not pass). Is this true explained this way?

app/src/androidTest/java/com/sdp/movemeet/view/navigation/NavigationTest.java
```diff
@@ -71,11 +71,6 @@
     @Before
     public void signIn(){
         Navigation.profileField = false;
```
Antho1426, line 73: Yes it seems like this technique simplifies a lot the process of "enabling the profile field of the user" or not! (Especially since now the same single variable is used everywhere in our activities)

app/src/main/java/com/sdp/movemeet/models/Activity.java
```diff
@@ -84,7 +84,7 @@ public Activity(String activityId,
         this.longitude = longitude;
         this.latitude = latitude;
         this.description = description;
-        super.documentPath = documentPath;
+        this.documentPath = documentPath;
```
Antho1426, line 87: Yes, I also updated this on my side! This allows us in fact to retrieve the documentPath of the activities without any problems! Using your `FirebaseObject` class as an abstract class rather than an interface would have been probably slightly cleaner, but in the end, I think going back to an interface is reasonable 👍.

app/src/main/java/com/sdp/movemeet/models/FirebaseObject.java
```diff
@@ -3,24 +3,18 @@
 /**
  * An object that can be found in the Firebase Realtime Database/Firestore backend
  */
-public abstract class FirebaseObject {
+public interface FirebaseObject {
```
Antho1426, line 6: Yes, going back to the "interface" version 👍. Thanks again for the time taken Monday morning to speak about this and try to find a solution to be able to retrieve the documentPath of the activities!

Antho1426 (approved):

PR #256, opened 2021-05-20
change order logic in onCreate to avoid useless objects

jlstoupy98:


app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
```diff
@@ -63,12 +63,7 @@
     @VisibleForTesting(otherwise=VisibleForTesting.PRIVATE)
     public static boolean enableNav = true;
 
-    DrawerLayout drawerLayout;
```
RoxaneBurri, line 66: Yes we don't need this lines !

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
```diff
@@ -90,6 +85,16 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_description);
 
+        fAuth = FirebaseAuth.getInstance();
+        if (fAuth.getCurrentUser() == null) {
```
RoxaneBurri, line 89: So you put the authentification first ?

RoxaneBurri (approved):

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
jlstoupy98: Yes exactly, I did that in all classes in such a way that it's the first thing that is done when we create the activity. It a bit more logic and it's avoid other useless objects creation.

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
Antho1426, line 66: Yes, in each of our activities, we don't need those DrawerLayout-related lines, thanks for removing them!

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
Antho1426: In fact, this is 1. a good necessary security check not to be left out and 2. this is now done in a logical way! Furthermore, it is now implemented in a consistent way across all our activities! Nice refactor, thanks for this Jean-Luc!

Antho1426 (approved):

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
jlstoupy98: Yeah, it's cleaner !

PR #271, opened 2021-05-21
Activity pictures cache

Walfar:
- Implemented cache for the activity header pictures
- Added class for permission checks
- tests implemented


app/src/androidTest/java/com/sdp/movemeet/utility/ActivityPictureCacheTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void cacheLoadsSamePictureAsSavedForSamePath()  {
+        activityRule.getScenario().onActivity(activity -> {
+            Bitmap bitmap = BitmapFactory.decodeResource(activity.getResources(), R.drawable.ic_add_black_24dp);
+            ActivityPictureCache.saveToCache(bitmap, "imagePath");
+            try {
+                Thread.sleep(5000);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+            Bitmap loadedBitmap = ActivityPictureCache.loadFromCache("imagePath");
```
Antho1426, line 43: Is this cache-dedicated class called "ActivityPictureCache" separated from the "ImageHandler" class as we discussed at the end of the previous sprint?

app/src/androidTest/java/com/sdp/movemeet/utility/ActivityPictureCacheTest.java
```diff
@@ -0,0 +1,47 @@
...
+            if (loadedBitmap != null) assertThat(loadedBitmap.sameAs(bitmap), is(true));
```
Antho1426, line 44: Ok ok this is the way you use to compare Bitmaps! On my side I had to implement an extra class called "DrawableMatcher" to perform such a comparison, that's interesting...

app/src/androidTest/java/com/sdp/movemeet/utility/LocationFetcherTest.java
```diff
@@ -51,7 +51,7 @@
             FragmentTestRule.create(MainMapFragment.class);
 
     @Rule
-    public GrantPermissionRule mRuntimePermissionRule = GrantPermissionRule.grant(Manifest.permission.ACCESS_FINE_LOCATION);
+    public GrantPermissionRule locationRule = GrantPermissionRule.grant(Manifest.permission.ACCESS_FINE_LOCATION);
```
Antho1426, line 54: Yes, variable naming convention. It's surprising, even if we refactored those names a few weeks ago, we always find variable names that are still unchanged! 😂

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -165,7 +164,7 @@ public void cameraZoomsOnlyOnFirstCall() {
     public void clickingOnActivityMakesCorrectIntentWhenLogged() {
         MainMapFragment mapFragment = fragmentTestRule.getFragment();
         Activity act = new Activity("activity id", "organizer id", "my title", 4, new ArrayList<>(), 0, 0,
-                    "desc", null, new Date(), 1, Soccer, "Dubai UAE", new Date());
+                    "desc", "activities/test", new Date(), 1, Soccer, "Dubai UAE", new Date());
```
Antho1426, line 167: Even if that's cleaner to set the documentPath of the activity this way, it was probably not necessary since this activity won't be actually uploaded to Firebase Firestore! But anyway, that's a good reflex, and it might help us figure out from where does our activity with a `null` documentPath get created at some point!

app/src/androidTest/java/com/sdp/movemeet/view/profile/EditProfileActivityImagePickerTest.java
```diff
@@ -87,7 +87,7 @@ public void uploadImageToFirebaseStorage() {
         ImageView profileImage = activityRule.getActivity().findViewById(R.id.image_view_edit_profile_image);
         Image image = new Image(imageUri, profileImage);
         image.setDocumentPath(userImagePath);
-        ImageHandler.uploadImage(image, progressBar);
+        ImageHandler.uploadImage(image,activityRule.getActivity());
```
Antho1426, line 90: Ok ok I see that you changed the signature of the method `uploadImage` of the `ImageHandler` class, I will analyse it down below 👍.

app/src/main/java/com/sdp/movemeet/utility/LocationFetcher.java
```diff
@@ -51,11 +54,12 @@ public LocationFetcher(SupportMapFragment supportMapFragment, LocationCallback l
     /**
      * Starts periodically requesting location updates to the fusedLocationProviderClient, if the permission is granted
      */
+    @SuppressLint("MissingPermission")
+    //Missing permission annotation because permissions are requested in isPermissionGranted() method
```
Antho1426, line 58: Ok, thank you for the comment!

app/src/main/java/com/sdp/movemeet/utility/PermissionChecker.java
```diff
@@ -0,0 +1,57 @@
+package com.sdp.movemeet.utility;
```
Antho1426, line 1: As we discussed, that's a very good point to have created a file gathering all the different permission checkers we need in our app, it makes our code more organized (we can now simply import the permissions in the activities requiring them)! Good job!

app/src/main/java/com/sdp/movemeet/utility/PermissionChecker.java
```diff
@@ -0,0 +1,57 @@
...
+     * Checks that permission for accessing GPS location is granted
+     * @return true if permission is granted, false otherwise
+     */
+    public static boolean isLocationPermissionGranted(Activity activity) {
+        return (ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
+                ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED);
+    }
+
+    public static boolean isStorageWritePermissionGranted(Activity activity) {
+        if (Build.VERSION.SDK_INT >= 23) {
```
Antho1426, line 25: Does it really mean that below Android SDK version 23, the user doesn't have to grant access to his storage (for writing and reading files) and this permission is granted by default? ^^'

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
```diff
@@ -315,9 +322,14 @@ private void loadActivityHeaderPicture() {
         imagePath = activity.getDocumentPath() + "/" + ACTIVITY_IMAGE_NAME;
         Image image = new Image(null, activityImage);
         image.setDocumentPath(imagePath);
-        ImageHandler.loadImage(image, progressBar);
+        ImageHandler.loadImage(image, this);
     }
 
+    public ProgressBar getProgressBar() {
```
Antho1426, line 440: Ok ok, that's what you said, since you changed the signature of the methods in the `ImageHandler` class which now take as argument the whole activity, you had to create a getter for the progress bar in `ActivityDescriptionActivity`, `ActivityDescriptionActivityUnregister`, `EditProfileActivity`, etc., I see 👍.

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -1,87 +1,118 @@
...
 
     private static BackendManager<Image> imageBackendManager;
 
+
     /**
      * Fetch an image (user profile picture, activity header picture or chat image) from Firebase Storage.
      *
      * @param image Image object to be loaded from Firebase Storage or from the local cache.
-     * @param progressBar A ProgressBar displayed as long as the image is not yet fetched.
+     * @param activity Activity on which we want to fetch the image
```
Antho1426, line 58: Here we can see the change you brought to the signature of the `loadImage` and `uploadImage` methods of the `ImageHandler` class 👍

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -1,87 +1,118 @@
...
      */
-    public static void loadImage(Image image, ProgressBar progressBar) {
-        if (progressBar != null) {
-            progressBar.setVisibility(View.VISIBLE);
```
Antho1426, line 43: Do you still handle the problem of the visibility of the progress bar somewhere in your updated class here?

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -1,87 +1,118 @@
...
+    public static void loadImage(Image image, Activity activity) {
+        Log.d(TAG, "loading image");
+        ProgressBar progressBar = getProgressBar(activity);
+        Bitmap bitmap = null;
+        ImageView imageView = image.getImageView();
+        String imagePath = image.getDocumentPath();
+
+        if (isStorageReadPermissionGranted(activity)) bitmap = loadFromCache(imagePath);
```
Antho1426, line 71: Ok, so the purpose is here to load the image from the cache in case there is an image to load, and otherwise to fetch the image from Firebase Storage. But did you think about the case where the image has been updated on Firebase Storage? In this case would your code manage to update the changed image or would it constantly load the same image previously saved in the cache?

app/src/main/java/com/sdp/movemeet/utility/ActivityPictureCache.java
```diff
@@ -0,0 +1,78 @@
...
+            out.close();
+            Log.d(TAG, "succcessfully cached image at path " + file.getPath());
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+    public static Bitmap loadFromCache(String path) {
+        Log.d(TAG, "Loading image from cache");
+        String imagePath = Environment.getExternalStorageDirectory().toString() + "/" + path;
```
Antho1426, line 77: So this location on the phone is not part of the actual app storage, isn't it? That's an external storage on the phone that is completely independent from the app location in the phone memory, is that correct? And if this is the case, I suppose the `path` argument should contain at least the word `movemeet/` at the beginning, no (to identify that this image file belongs to the movemeet app)?

app/src/androidTest/java/com/sdp/movemeet/utility/ActivityPictureCacheTest.java
Walfar: Yes, you can simply use this method, and it works on local. But, for some reason, on Cirrus, the loaded bitmap is null, implying that the bitmap was not stored. For the moment, I simply added a check, but it would be better to find a way to make this test pass with Cirrus.

app/src/androidTest/java/com/sdp/movemeet/utility/LocationFetcherTest.java
Walfar: indeed haha

app/src/androidTest/java/com/sdp/movemeet/utility/ActivityPictureCacheTest.java
Walfar: Yes, although it is used in the ImageHandler. I thought it would be clearer this way, and easier to test.

Walfar: Yes, although it is used in the ImageHandler. I thought it would be clearer this way, and easier to test.

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
Walfar: True. I changed it because I thought the null documentPath bug in our db came from that, but as you stated, this activity isn't even uploaded to Firebase.

app/src/main/java/com/sdp/movemeet/view/activity/ActivityDescriptionActivity.java
Walfar: Exact ! It is necessary to have it for the permission checker, so I can directly access all the elements needed with getters. But, there is still a bug with  the progress bar that doesn't seem to always appear...

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Walfar: Good catch ! I didn't set back the visibility of the progressbar... I will change that immediately !

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Walfar: It would indeed always load the same image. I will try to change that next week with a listener on the database

app/src/main/java/com/sdp/movemeet/utility/ActivityPictureCache.java
Walfar: You,re right ! I made this dummy path, to see if the caching worked. But now that I know it works, it would be wise to set a clear path for the cache

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Antho1426: Yes, thank you very much!

Antho1426 (approved):

app/src/main/java/com/sdp/movemeet/utility/ActivityPictureCache.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public static String TAG = "Cache TAG";
+
+    /**
+     * Saves a bitmap to the local cache, for the given path
+     * @param bitmap bitmap to cache
+     * @param path path where we save the bitmap
+     */
+    public static void saveToCache(Bitmap bitmap, String path) {
+        String[] directories = path.split("/");
```
Antho1426, line 44: Just in case, I have created a constant for this "/" symbol we kind of use everywhere in the app. It's in ImageHandler and is called "PATH_SEPARATOR".

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -1,96 +1,161 @@
...
+import static com.sdp.movemeet.utility.PermissionChecker.isStorageReadPermissionGranted;
+import static com.sdp.movemeet.utility.PermissionChecker.isStorageWritePermissionGranted;
 
 /**
  * This class allows to both load images from and upload images to the Firebase Storage service.
  * In case the image is not present in the local cache, it is fetched from Firebase Storage.
  * On the contrary, if the the image is already in the local cache, it is simply loaded from there.
  */
-public class ImageHandler {
+public abstract class ImageHandler {
```
Antho1426, line 41: Why did you change this class to an abstract class, is it necessary?

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
```diff
@@ -1,96 +1,161 @@
...
+
+        if (bitmap != null) {
+            // bitmap was successfully loaded from cache
+            Log.d(TAG, "bitmap stored");
+            imageView.setImageBitmap(bitmap);
+            if (progressBar != null) {
+                progressBar.setVisibility(View.GONE);
+            }
+        }
+        //Then load from database
```
Antho1426, line 81: Correct, the process is following:
1.	In any case, we first load the image from the cache if it’s available
2.   	Then, we can check if a new image is available in the database and if this is the case, we update the picture!

app/src/main/java/com/sdp/movemeet/view/map/MainMapFragment.java
```diff
@@ -110,8 +112,8 @@ public void onLocationResult(@NonNull LocationResult locationResult) {
 
         locationFetcher = new LocationFetcher(supportMapFragment, locationCallback);
 
-        //We also define the main OnMapReady callback for setting the map listeners
-        supportMapFragment.getMapAsync(this);
+        //Update the local list of activities from the database. On success, we update the map by dispalying the activities markers
+        updateListActivities(o -> displayNearbyMarkers());
```
Antho1426, line 116: So, if I'm not mistaken, this is the trick we discussed last week where you use to automatically refresh the map with the updated activities each time the MainMapFragment is accessed, correct?

Antho1426 (approved):

app/src/main/java/com/sdp/movemeet/utility/ActivityPictureCache.java
Walfar: good to know, thanks ! I'll use it

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Walfar: It's just to specify that this class shouldn't be instanciated, as all its methods are static anyway. But it doesn't change anything to the functionality of the code.

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Walfar: Exactly ! No matter what, there will be traffic with the db, but the image will load faster as it's first loaded from cache (and will appear even without internet connection). We always have to fetch from db, in the case where there already is an image in the cache, but the image has been changed by the organizer.

app/src/main/java/com/sdp/movemeet/view/map/MainMapFragment.java
Walfar: This update is called at the creation of the fragment, so this is not the fix. But there is a similar call on resume, that first removes the markers and then re-updates the list. When we refresh the map, e.g going on another activity and coming back to the map fragment, the fragment is resumed (as the activity was not destroyed).

app/src/main/java/com/sdp/movemeet/utility/ImageHandler.java
Antho1426: Ok ok I see, thank you !

app/src/main/java/com/sdp/movemeet/view/map/MainMapFragment.java
Antho1426: Yes, good trick! This way, the updates from ActivityDescriptionActivity to Firebase Firestore inevitably be taken into account ! Not sure if we'll have time, but using a listener instead would probably be a more optimized method 👌 as we discussed. Anyway, we have at least a working solution now 😉.

app/src/main/java/com/sdp/movemeet/utility/ActivityPictureCache.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.storage.StorageTask;
+import com.sdp.movemeet.backend.firebase.storage.StorageImageManager;
+import com.sdp.movemeet.models.Image;
+import com.sdp.movemeet.view.activity.ActivityDescriptionActivity;
+import com.squareup.picasso.Picasso;
+
+import java.io.File;
+import java.io.FileOutputStream;
+
```
RoxaneBurri, line 30: You can add a comment in the beginning for explain the goal of this class

app/src/main/java/com/sdp/movemeet/utility/PermissionChecker.java
RoxaneBurri: +1

app/src/main/java/com/sdp/movemeet/view/chat/MessageViewHolder.java
```diff
@@ -79,12 +79,12 @@ public void bindMessage(Message message, ChatActivity chatActivity) {
      * @param message Message object containing the message data (in this case the message data of
      *                interest is the URL of the image)
      */
-    private void handlingUserProfilePicture(Message message) {
+    private void handlingUserProfilePicture(Message message, ChatActivity chatActivity) {
```
RoxaneBurri, line 82: Why the chatActivity is add in this function ?

app/src/main/java/com/sdp/movemeet/view/home/HomeScreenActivity.java
```diff
@@ -33,11 +33,6 @@ protected void onCreate(Bundle savedInstanceState) {
         if (ActivityCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) != PackageManager.PERMISSION_GRANTED) {
             ActivityCompat.requestPermissions(this, new String[]{Manifest.permission.ACCESS_FINE_LOCATION}, REQUEST_CODE);
         }
-
```
RoxaneBurri, line 36: Yes we really have to remove useless code

RoxaneBurri (approved):

app/src/main/java/com/sdp/movemeet/utility/ActivityPictureCache.java
Walfar: Indeed !

app/src/main/java/com/sdp/movemeet/view/chat/MessageViewHolder.java
Walfar: It's necessary as the image handler uses the activity it loads the image to (we need the context to check the permissions and get access to the progress bar).

app/src/main/java/com/sdp/movemeet/view/home/HomeScreenActivity.java
Walfar: Yep, we will have to do a last round of check, to make sure that there are no more of these.

PR #277, opened 2021-05-27
change user constructor

RoxaneBurri:


app/src/main/java/com/sdp/movemeet/backend/serialization/UserSerializer.java
```diff
@@ -15,6 +16,10 @@
     public static final String PHONE_KEY = "phone";
     // The key used to access the description attribute of a serialized User
     public static final String DESCRIPTION_KEY = "description";
+    // The key used to access the created activities attribute of a serialized User
+    private static  final String CREATED_ACTIVITIES_KEY = "createdActivity";
```
Antho1426, line 20: Nice! Thank you for taking our conversation into account and adding this field in the `User` objects! This way, as suggested a few weeks ago by @maximilian1064, we will be able to not only completely delete the user from Firebase but also the "Firebase Firestore documents" of the activities he created as an organiser! The "delete account" feature will hence make much more sense!

app/src/main/java/com/sdp/movemeet/models/User.java
```diff
@@ -17,40 +16,45 @@
     private String idUser;
     private String imageId;
     private String documentPath;
+    private ArrayList<String> createActivitiesId;
```
Antho1426: Could you just please rename the name of this ArrayList from createActivitiesId to createdActivitiesId? I think that the missing "d" is a typo here...

app/src/main/java/com/sdp/movemeet/models/User.java
```diff
@@ -111,11 +118,25 @@ public String getDocumentPath() {
         return this.documentPath;
     }
 
+    /**
+     * @return the list of activities created by the user
+     */
+    public ArrayList<String> getCreateActivitiesId() {
```
Antho1426: Here as well, I would rather name this getter "getCreatedActivitiesId" (with a "d").

app/src/main/java/com/sdp/movemeet/models/User.java
RoxaneBurri: Yes I did it in my last commit !

bitshrine (approved):

app/src/main/java/com/sdp/movemeet/models/User.java
Antho1426: Nice thank you!

Antho1426 (approved):

PR #292, opened 2021-06-04
Fixed GPS units

bitshrine:


app/src/main/java/com/sdp/movemeet/models/GPSPath.java
```diff
@@ -141,7 +141,7 @@ public static float computeAverageSpeed(float distance, long time) {
         if (time <= 0) throw new IllegalArgumentException();
         if (distance <= 0) return 0;
 
-        return (distance / 1000) / (time / 3600);
+        return (distance * 1000) / (time / 3600.0f);
```
jlstoupy98, line 144: Why this multiplication by 1000 rather than a division of the time by 3600 * 1000 ?

jlstoupy98 (approved):

app/src/main/java/com/sdp/movemeet/models/GPSPath.java
bitshrine: Dividing it by 3.6e+6 often caused the `time` value to be completely cancelled into 0, meaning that it would then lead to a bad value for the average speed. Expressing the equation this way allows the values to be computed correctly

Walfar (approved):
Nothing to add from what JL said. Good job !

PR #293, opened 2021-06-04
Optimize the update of activities

Walfar:
Instead of clearing all markers on map and refetching all activities from DB, we listen for changes in the DB, and only update the marker corresponding to the activity that changed.
- Added a snapshot listener on document in FirestoreActivityManager
- Added a HashMap associating an activity id to a marker, to pick the exact activity that was updated and remove/add the corresponding marker
- When fetching an activity: add a listener to the doc, then put the activity and its marker in the HashMap
- When changes is detected, remove marker and add a new one with a tag associated to the updated activity
- Added permission checks for gallery intent when editing picture


app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
```diff
@@ -171,7 +171,7 @@ public void clickingOnActivityMakesCorrectIntentWhenLogged() {
             public void run() {
                 mapFragment.displayNearbyMarkers();
                 if (mapFragment.googleMap != null) {
-                    mapFragment.onMarkerClick(mapFragment.activitiesMarkers.get(0));
+                    mapFragment.onMarkerClick(mapFragment.activitiesMarkerMap.get(act));
```
Antho1426, line 174: Is this change (.get(0) --> .get(act)) related to your optimisation for fetching *only* the updated activity-ies?

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
```diff
@@ -70,11 +74,13 @@ public static void updateListActivities(OnCompleteListener listener) {
      *
      * @param task task containing the requested snapshots, that would then be deserialized into activities
      */
-    private static Task addActivitiesOnSuccess(Task<QuerySnapshot> task) {
+    private static Task addActivitiesOnSuccess(Task<QuerySnapshot> task, EventListener<DocumentSnapshot> eventListener) {
```
Antho1426, line 77: I guess this is the place where you implemented your listener to listen to updated activities. Good that you had the time to implement this!

app/src/main/java/com/sdp/movemeet/view/profile/EditProfileActivity.java
```diff
@@ -109,8 +111,10 @@ private void assignViewsAndAdjustData() {
 
 
     public void changeProfilePicture(View view) {
-        Intent openGalleryIntent = new Intent(Intent.ACTION_PICK, MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
-        startActivityForResult(openGalleryIntent, REQUEST_IMAGE);
+        if (isStorageReadPermissionGranted(this)) {
```
Antho1426, line 130: Nice, thanks for adding this! As discussed, otherwise we could access the gallery to select a profile picture by default, without any permission...

Antho1426 (approved):

app/src/androidTest/java/com/sdp/movemeet/view/map/MainMapFragmentTest.java
Walfar: At first, I implemented a list of markers in the map fragment, only in the interest of making this test work (as I can't have access to a marker directly from the map). But, with the new implementation, I had to associate each activities with their respective markers, so that I could only update the marker of the activity that changed. Doing that as a HashMap, I figured that I could directly use this map instead of a list in this test, killing two birds with one stone.

Walfar: At first, I implemented a list of markers in the map fragment, only in the interest of making this test work (as I can't have access to a marker directly from the map). But, with the new implementation, I had to associate each activities with their respective markers, so that I could only update the marker of the activity that changed. Doing that as a HashMap, I figured that I could directly use this map instead of a list in this test, killing two birds with one stone.

app/src/main/java/com/sdp/movemeet/utility/ActivitiesUpdater.java
Walfar: Yes ! Maybe this isn't the smoothest way to do it, but I didn't want to change too much of the already implemented architecture.

app/src/main/java/com/sdp/movemeet/view/profile/EditProfileActivity.java
Walfar: True, although I still find it weird that android studio didn't complain about the version without checks...

app/src/main/java/com/sdp/movemeet/utility/ActivityPictureCache.java
```diff
@@ -28,9 +28,6 @@
 import java.io.File;
 import java.io.FileOutputStream;
 
-/**
- * This utility class is used to load/save the activities header pictures from/to cache
- */
```
bitshrine, line 33: Why did you delete this bit of documentation?

bitshrine (approved):

